PR #45, opened 2022-03-04
Implement the authentication API with Firebase auth

alexandrepiveteau:
This PR implements the authentication API with Firebase authentication, and adds some associated tests.

Additionally, the main screen of the application is updated to display the authentication UI by default.


mobile/src/main/java/ch/epfl/sdp/mobile/data/api/firebase/FirebaseAuthenticationApi.kt
```diff
@@ -0,0 +1,86 @@
...
+/**
+ * An implementation of the [AuthenticationApi] which uses [FirebaseAuth] internally.
+ *
+ * @param auth the [FirebaseAuth] instance which will be used to handle authentication.
+ */
+class FirebaseAuthenticationApi(private val auth: FirebaseAuth) : AuthenticationApi {
+
+  override val currentUser: Flow<AuthenticationApi.User> =
+      auth.currentUserFlow()
+          .map { user ->
```
BadrTad, line 24: [Proposition] It took me some time to understand what was going on here (maybe because I am new to Kotlin) However I suggest this 😃.  What do you think ?  
```suggestion
          .mapNotNull { user ->  FirebaseUserAdapter(it, auth)}
          .onEmpty { AuthenticationApi.User.NotAuthenticated }
          .onStart { emit(AuthenticationApi.User.Loading) }
```

mobile/src/androidTest/java/ch/epfl/sdp/mobile/data/features/authentication/api/firebase/FirebaseAuthenticationApiTest.kt
```diff
@@ -0,0 +1,162 @@
...
+import com.google.firebase.auth.FirebaseUser
+import io.mockk.every
+import io.mockk.mockk
+import io.mockk.verify
+import kotlinx.coroutines.flow.filterIsInstance
+import kotlinx.coroutines.flow.first
+import kotlinx.coroutines.test.runTest
+import org.junit.Test
+
+class FirebaseAuthenticationApiTest {
```
BadrTad, line 21: Tests looks good :+1:

mobile/src/main/java/ch/epfl/sdp/mobile/data/api/firebase/FirebaseAuthenticationApi.kt
```diff
@@ -0,0 +1,86 @@
...
+ *
+ * @receiver the [FirebaseAuth] from which the current user is retrieved.
+ */
+private fun FirebaseAuth.currentUserFlow(): Flow<FirebaseUser?> =
+    callbackFlow {
+          val listener = FirebaseAuth.AuthStateListener { auth -> trySend(auth.currentUser) }
+          addAuthStateListener(listener)
+          awaitClose { removeAuthStateListener(listener) }
+        }
+        .buffer(capacity = 1, onBufferOverflow = BufferOverflow.DROP_OLDEST)
```
BadrTad, line 70: [Question] Why do we buffer only one user at a time ?

BadrTad (approved):
Great work ! I learnt a lot about Flows, Coroutines and Kotlin as I was reviewing your PR

mobile/src/main/java/ch/epfl/sdp/mobile/data/api/firebase/FirebaseAuthenticationApi.kt
alexandrepiveteau: I think the semantics of this suggested change are slightly different than what I had in mind - in particular, the `mapNotNull` operator will filter out `null` values in the resulting Flow, when we want to preserve them. Also, `onEmpty` will simply let the `Flow` emit certain values if it's empty, but won't be used otherwise. It's a good way to provide a "default value" to Flows which might emit 0 elements.

Just fyi, when using an operator like `onEmpty {}`, you should be cautious to call `emit` if you want to emit an element; otherwise, the statement is executed but won't issue items in the Flow.

Your concerns about the readability of the code are valid though. I'll extract this code to a dedicated function to make it easier to grasp.

mobile/src/main/java/ch/epfl/sdp/mobile/data/api/firebase/FirebaseAuthenticationApi.kt
alexandrepiveteau: We preserve only the last value of the Flow, and we can coalesce values (since generally, only the current value of the current user is interesting). This capacity of 1 plays well with the `trySend` operator, because we guarantee that `trySend` will succeed and that flow collectors will always have the "most up-to-date" user.

mobile/src/main/java/ch/epfl/sdp/mobile/data/api/firebase/FirebaseAuthenticationApi.kt
alexandrepiveteau: Done in commit 17000705893bc72125209f43936ff838d265c276 😄

mobile/src/main/java/ch/epfl/sdp/mobile/HomeActivity.kt
```diff
@@ -3,14 +3,34 @@ package ch.epfl.sdp.mobile
...
 class HomeActivity : ComponentActivity() {
 
   override fun onCreate(savedInstanceState: Bundle?) {
     super.onCreate(savedInstanceState)
-    setContent { ProvideLocalizedStrings { Text("Hello world") } }
+
+    val api = FirebaseAuthenticationApi(Firebase.auth)
+
+    setContent {
+      ProvideLocalizedStrings {
```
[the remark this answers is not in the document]
alexandrepiveteau: The localised strings interface offers a way to easily swap the strings used in the app at runtime (which may prove to be useful if we want to offer a language picker in the settings). Compose offers much more control over the configuration than what you'd get with the standard Android resources, so this dedicated interface tries to leverage that flexibility to offer more flexibility when it comes to choosing which strings we use.

I think we can split in multiple smaller interfaces if it becomes too big.

It's a pattern I've used in two other apps already, it has some benefits (it makes some testing easier sometimes) but also some drawbacks. Maybe it's something we could discuss with the rest of the team soon !

BadrTad (dismissed):

mobile/src/androidTest/java/ch/epfl/sdp/mobile/data/features/authentication/api/firebase/FirebaseAuthenticationApiTest.kt
```diff
@@ -0,0 +1,162 @@
...
+    every { auth.createUserWithEmailAndPassword(any(), any()) } returns
+        Tasks.forException(Exception())
+
+    val api = FirebaseAuthenticationApi(auth)
+
+    assertThat(api.signUpWithEmail("email", "password")).isEqualTo(Failure)
+  }
+
+  @Test
+  fun signUp_withWithoutAuthResult_returnsFailure() = runTest {
```
matt989253, line 85:
```suggestion
  fun signUp_withoutAuthResult_returnsFailure() = runTest {
```

matt989253: Small typo, apart from that, LGTM!

matt989253 (approved):

BadrTad (approved):

PR #47, opened 2022-03-05
Implement authentication navigation and bottom navigation

alexandrepiveteau:
This pull request implements basic navigation, using the AuthenticationApi to determine if the user is currently logged in, and using the AndroidX navigation component once the user is properly logged in.

Placeholders for the stateful `FollowingScreen` and `SettingsScreen` composables are provided.

This PR also introduces `AlwaysSucceedingAuthenticationApi` in the test suite, which can be used to simulate an `AuthenticationApi` which always succeeds.


matt989253 (dismissed):
LGTM, this should allow the others to work on their tasks efficiently!

mobile/src/main/java/ch/epfl/sdp/mobile/ui/Api.kt
```diff
@@ -0,0 +1,25 @@
...
+val LocalAuthenticationApi = compositionLocalOf<AuthenticationApi> { error("Missing API.") }
+
+/**
+ * Provides the given APIs through different [androidx.compose.runtime.CompositionLocal] values
+ * available throughout the hierarchy.
+ *
+ * @param authentication the [AuthenticationApi] that will be provided.
+ */
+@Composable
+fun ProvideApis(
```
KurohanaJuri, line 18: Suggestion : maybe change the filename to a more specific one that `Api.kt`?

mobile/src/main/java/ch/epfl/sdp/mobile/ui/features/home/StatefulHome.kt
```diff
@@ -0,0 +1,66 @@
...
+import androidx.navigation.compose.NavHost
+import androidx.navigation.compose.composable
+import androidx.navigation.compose.currentBackStackEntryAsState
+import androidx.navigation.compose.rememberNavController
+import ch.epfl.sdp.mobile.data.api.AuthenticationApi
+import ch.epfl.sdp.mobile.ui.features.profile.StatefulProfileScreen
+import ch.epfl.sdp.mobile.ui.features.social.StatefulFollowingScreen
+
+/** The route associated to the social tab. */
+private const val SocialRoute = "social"
```
KurohanaJuri, line 18: Modification : variable name should be uppercase

```suggestion
private const val SOCIAL_ROUTE= "social"
```

mobile/src/main/java/ch/epfl/sdp/mobile/ui/features/home/StatefulHome.kt
```diff
@@ -0,0 +1,66 @@
...
+
+/** The route associated to the settings tab. */
+private const val SettingsRoute = "settings"
```
KurohanaJuri, line 21: Modification : variable name should be uppercase

```suggestion
private const val SETTINGS_ROUTE = "settings"
```

mobile/src/main/java/ch/epfl/sdp/mobile/ui/Api.kt
alexandrepiveteau: Sure thing, I'll rename it to `ApiCompositionLocals.kt`, to better reflect that the code only relates to providing some composition locals for the different `***Api.kt` we will have.

mobile/src/main/java/ch/epfl/sdp/mobile/ui/features/home/StatefulHome.kt
alexandrepiveteau: I'm not sure about that one, since the AndroidX repository seems to accept both naming conventions for internal / private usage (see https://github.com/androidx/androidx/search?l=Kotlin&q=const).

However, all of the public Compose API seems to use PascalCase, since it tries to avoid distinguishing between `public const val` and `const val` values. To keep it consistent with the Compose API, I'd suggest we keep PascalCase for now, but we can discuss this with the team in the next meeting. Does this sound good to you ?

mobile/src/main/java/ch/epfl/sdp/mobile/ui/features/home/StatefulHome.kt
alexandrepiveteau: (same as above)

mobile/src/main/java/ch/epfl/sdp/mobile/ui/Api.kt
alexandrepiveteau: Done in 2dbf458370d3705d49308c8a47038ef50fec9c52.

matt989253 (approved):

KurohanaJuri (dismissed):

BadrTad (changes requested):
Good work !

mobile/src/main/java/ch/epfl/sdp/mobile/ui/features/Navigation.kt
```diff
@@ -0,0 +1,64 @@
...
+    when (it) {
+      is UserState.Authenticated -> StatefulHome(it.user)
+      UserState.NotAuthenticated -> StatefulAuthenticationScreen()
+      UserState.Loading -> Unit
+    }
+  }
+}
+
+/** An interface representing the current state of the user. */
+private sealed interface UserState {
```
BadrTad, line 29: I think that this interface deserves it's own file. Could this `UserState` interface be used throughout the whole project, for `ProfileState` for instance.

mobile/build.gradle.kts
```diff
@@ -61,6 +61,7 @@ dependencies {
   testImplementation(libs.truth)
   testImplementation(libs.mockk.mockk)
   testImplementation(libs.mockk.agent.jvm)
+  androidTestImplementation(libs.junit4)
```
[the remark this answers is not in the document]
alexandrepiveteau: Not by default. Some third-party integrations (like https://github.com/mannodermaus/android-junit5) offer Junit 5 support for Android, however Junit4 + Google Truth offer a solid first-party alternative with a good assertion framework.

Also, only Junit 4 is supported for UI testing with Compose (https://developer.android.com/jetpack/compose/testing#setup).

mobile/src/androidTest/java/ch/epfl/sdp/mobile/data/features/authentication/AlwaysSucceedingAuthenticationApi.kt
```diff
@@ -0,0 +1,48 @@
...
+import ch.epfl.sdp.mobile.data.api.AuthenticationApi
+import ch.epfl.sdp.mobile.data.api.AuthenticationApi.AuthenticationResult.Success
+import ch.epfl.sdp.mobile.data.api.AuthenticationApi.User
+import kotlinx.coroutines.flow.Flow
+import kotlinx.coroutines.flow.MutableStateFlow
+import kotlinx.coroutines.flow.map
+
+/**
+ * A fake implementation of [AuthenticationApi] which is backed by an internal state flow, and whose
+ * calls too [signInWithEmail] and [signUpWithEmail] will always succeed.
```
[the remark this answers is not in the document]
alexandrepiveteau: Fixed in f2a3f517ae41c98980c006b9f9bb8cd536f82fc9.

mobile/src/main/java/ch/epfl/sdp/mobile/ui/features/Navigation.kt
alexandrepiveteau: I'm not sure about that. I made the `UserState` interface private to this file, since it's not used anywhere else (and I don't think it will be); it's here only to model nicely how the `AuthenticationApi` used by the `Navigation` composable is selects which stateful composable should be displayed (nothing aka. splash screen, auth screen or home).

The `ProfileState` will always be used when the user is authenticated, so a `UserState` modelling if the user is logged in won't help.

matt989253 (approved):

mobile/src/main/java/ch/epfl/sdp/mobile/ui/features/Navigation.kt
BadrTad: Clear !

mobile/src/androidTest/java/ch/epfl/sdp/mobile/data/features/authentication/AlwaysSucceedingAuthenticationApi.kt
```diff
@@ -0,0 +1,48 @@
...
+import ch.epfl.sdp.mobile.data.api.AuthenticationApi.User
+import kotlinx.coroutines.flow.Flow
+import kotlinx.coroutines.flow.MutableStateFlow
+import kotlinx.coroutines.flow.map
+
+/**
+ * A fake implementation of [AuthenticationApi] which is backed by an internal state flow, and whose
+ * calls to [signInWithEmail] and [signUpWithEmail] will always succeed.
+ */
+class AlwaysSucceedingAuthenticationApi : AuthenticationApi {
```
BadrTad, line 14: Why do we need an AlwaysSucceedingAuthenticationApi fake API ?

mobile/src/main/java/ch/epfl/sdp/mobile/ui/i18n/English.kt
```diff
@@ -10,4 +10,7 @@ object English : LocalizedStrings {
   override val authenticateSwitchToLogIn = "Log in with email"
   override val authenticateSwitchToRegister = "Register with email"
   override val authenticateErrorFailure = "Something went wrong when authenticating"
+
+  override val sectionSocial = "Players"
+  override val sectionSettings = "Settings"
```
BadrTad, line 15: Do we actually specify what language to use in the app ? Or is it implicitly done ?

mobile/src/androidTest/java/ch/epfl/sdp/mobile/data/features/authentication/AlwaysSucceedingAuthenticationApi.kt
alexandrepiveteau: We definitely could, but having a fake API implementation offers some flexibility when writing certains tests too. Is there something specific that would make mocks a more obvious choice ?

mobile/src/main/java/ch/epfl/sdp/mobile/ui/i18n/English.kt
alexandrepiveteau: We don't have more than English yet, but we have all the infrastructure required for languages that change based on device configuration (see `LocalLocalizedStrings`).

BadrTad: 👍

mobile/src/androidTest/java/ch/epfl/sdp/mobile/data/features/authentication/AlwaysSucceedingAuthenticationApi.kt
BadrTad: Not really, I was just wondering about the reason that made you create a fake API over mocking 😀

BadrTad (approved):
Thanks for the clarifications 😃

PR #48, opened 2022-03-06
Add a mocking layer for Firestore

alexandrepiveteau:
Following some discussions with @matt989253 and @barmettlerl, it seems like a mocking layer very close to the Firestore API might be needed to efficiently write some tests for the implementation of the Api layer.

This PR attempts to provide such a low-level mocking layer, with an actual implementation which delegates to Firestore and a fake implementation for testing.

---

Example usage of the fake implementation :

```kotlin
suspend fun main() {
  val store = buildFakeStore {
    collection("users") {
      document("alexandre", null)
      document("matthieu", "document_matthieu")
      document("lars", "document_lars")
    }
  }

  // Prints document_matthieu
  store.collection("users").document("matthieu").asFlow<String>().take(1).collect(::println)

  // Prints null, document_matthieu, document_lars
  store.collection("users").asFlow<String>().take(1).collect { println(it.joinToString()) }
}
```

More usage samples can be found in the tests of the mocking layer, in the file `DslTest.kt`.


mobile/src/androidTest/java/ch/epfl/sdp/mobile/backend/store/fake/DslTest.kt
```diff
@@ -0,0 +1,106 @@
...
+      this["title"] = "Hello"
+      this["subtitle"] = "World"
+    }
+    store.collection("users").document("doc").set { this["title"] = "Hello" }
+
+    val data = store.collection("users").document("doc").asFlow<SampleDocument>().first()
+
+    assertThat(data).isNotNull()
+    assertThat(requireNotNull(data).subtitle).isNull()
+  }
```
matt989253, line 89: Nitpick: Maybe this test should also test that the set value is the same as the one later retrieved?

matt989253 (commented):
Looks very good (and impressive) to me. I don't feel really qualified enough to comment on the DSL implementation, but the tests look quite exhaustive so i'm not worried.

mobile/src/androidTest/java/ch/epfl/sdp/mobile/backend/store/fake/DslTest.kt
alexandrepiveteau: Done in 15bca107a26f796f60c7897a84436d191fea0b86.

matt989253 (approved):

mobile/build.gradle.kts
```diff
@@ -67,6 +67,7 @@ dependencies {
   androidTestImplementation(libs.androidx.test.truth)
   androidTestImplementation(libs.androidx.test.espresso)
   androidTestImplementation(libs.compose.ui.test.junit4)
+  androidTestImplementation(libs.kotlin.reflect)
```
barmettlerl, line 70: Comment: I do not really understand why we need to add the library reflect to the tests implementation. Do we run the tests for the library or do we provide this library for our tests?

barmettlerl:
Looks also good for me.

mobile/build.gradle.kts
alexandrepiveteau: The library is an interface which defines how to get data from Firestore.

An implementation which uses the actual Firestore SDK is provided in the application, and is tested with standard mocking. Also, a mock implementation is provided in the androidTest source set, which stores data in memory and makes use of reflection.

barmettlerl (approved):

PR #57, opened 2022-03-08
Feature/lb/all people viewmodel

barmettlerl:


mobile/src/main/java/ch/epfl/sdp/mobile/data/api/firebase/FirebaseAuthenticatedUser.kt
```diff
@@ -52,7 +53,7 @@ class FirebaseAuthenticatedUser(
       updated["name"] = scope.name
     }
     return try {
-      firestore.collection("users").document(user.uid).set(updated, SetOptions.merge()).await()
+      user?.uid?.let { firestore.collection("users").document(it).set(updated, SetOptions.merge()).await() }
```
alexandrepiveteau, line 56: Bug : you may end up querying `firestore.collection("users").document("null")` if the `auth.currentUser` is `null`.

PR #62, opened 2022-03-09
Finalize the authentication screen

Fouad-sys:
-Designed the sign up screen
-Adjusted the designs of the login and signup screens to resemble their figma mockups
-Added the name field for the sign up screen
-Animated the transition between the login and signup screens
-Modified the implementation of the authentication sign up page


alexandrepiveteau (approved):
Looks good to me, thanks !

KurohanaJuri (approved):
It's looks great, but I think we can improve some aspects :
- When the user sign up, ensure that the given email has at least the form `email@email`
- For small screen, it would be great that screen scroll up if the text field is hidden by the keyboard

PR #87, opened 2022-03-11
Implement authentication mocking

alexandrepiveteau:
This PR adds some `Auth` and `User` interfaces, which can be used to mock Firebase authentication. Additionally, some fake in-memory implementations are provided, alongside with a simple DSL to build an `Auth` instance with some users.

Some additional work is still require to migrate the `AuthenticationFacade` interface to use this `Auth` interface, and will come in some follow-up PRs.

---

## DSL Example

```kotlin
val auth = buildAuth {
  user("alexandre.piveteau@epfl.ch", "password")
  user("john.cena@epfl.ch", "HELLO")
}

val result = auth.signInWithEmail("alexandre.piveteau@epfl.ch", "password") // Success
```


KurohanaJuri (approved):
Looks good to me

Fouad-sys (approved):
Nice mock!

PR #89, opened 2022-03-11
Refactor the authentication facade to use the `Auth` mockable layer

alexandrepiveteau:
This PR builds on top of #89 to refactor the `AuthenticationFacade` and make it use `Auth` rather than `FirebaseAuth`. Additionally, it gets rid of the `ProfileColor` enum and replaces it with `Profile.Color`, which models colors as their hexadecimal code.


Fouad-sys (approved):
Nicely done.

barmettlerl (approved):

PR #90, opened 2022-03-13
Introduce `DocumentSnapshot` and `QuerySnapshot` to stores

alexandrepiveteau:
This PR introduces the `DocumentSnapshot` and `QuerySnapshot` interfaces (from the Firestore API) to the store abstraction. Until now, `Query.asFlow()` and `DocumentReference.asFlow()` would return some typed results. However, this would make implementing some field operators, such as `whereEquals(field: String, value: Any?)`, particularly difficult, since the operators would have to work on a generic `Flow<List<T>>`. The introduction of `DocumentSnapshot` and `QuerySnapshot` will help simplify the (future) implementation of these operators for the `FakeStore`.

The following changes are included in the PR :

- `toObject()` and `fromObject()` have been moved to a separate package;
- `DocumentSnapshot` and `QuerySnapshot` have been introduced;
- the implementation of `Query.asFlow()` has been reworked and fixes the existing generic type erasure issue; and
- `AbstractQuery` has been renamed to `FakeQuery` and made an interface.


barmettlerl (approved):
Looks good for me. 👌

Fouad-sys (approved):
Awesome simplification, thanks!

PR #93, opened 2022-03-13
Add support for `whereEquals`, `whereNotEquals` and `whereArrayContains` to `Query`

alexandrepiveteau:
The fake implementation is less flexible than the real Firestore, because it only supports `List` types in documents, but that's probably good enough to implement user following and search.


barmettlerl (approved):

Fouad-sys (approved):
Awesome thanks!

PR #96, opened 2022-03-14
Add `PieceIdentifier` to pieces

alexandrepiveteau:
This PR introduces `PieceIdentifier` to `Piece`, which acts as a disambiguator when two pieces share the same color and rank. A disambiguator is necessary to implement smooth UI transitions between different board states, to keep track of how each piece should animate within a move.

While I'm not fully convinced by the API design for the identifiers right now (especially because they seem like they're an abstraction leak of how `Piece` is defined), it's probably better to ship them as-is for now, so @matt989253 can use it when implementing the `Board` composable.


Fouad-sys (approved):
Looks good

matt989253 (approved):

PR #97, opened 2022-03-14
Add a chessboard screen

matt989253:


alexandrepiveteau (approved):

mobile/src/main/java/ch/epfl/sdp/mobile/ui/game/ChessBoardState.kt
```diff
@@ -0,0 +1,68 @@
...
+    Bishop,
+    Knight,
+    Pawn,
+  }
+
+  /** The different colors which may be displayed by a [ChessBoard]. */
+  enum class Color {
+    Black,
+    White,
+  }
```
KurohanaJuri, line 32: Why can't we use the enum that was already written [(this)](https://github.com/epfl-SDP/android/blob/main/mobile/src/main/java/ch/epfl/sdp/mobile/application/chess/Rank.kt) ?

mobile/src/main/java/ch/epfl/sdp/mobile/ui/game/ChessBoard.kt
```diff
@@ -0,0 +1,203 @@
...
+      modifier
+          .aspectRatio(1f)
+          .checkerboard(
+              cells = 8,
+              color = MaterialTheme.colors.onPrimary.copy(alpha = ContentAlpha.disabled),
+          )
+          .grid(cells = 8, color = MaterialTheme.colors.primary),
+  ) {
+    val minDimension = with(LocalDensity.current) { min(maxHeight, maxWidth).toPx() }
+    val cellPx = minDimension / 8
```
KurohanaJuri, line 52: /!\ Magic number

mobile/src/main/java/ch/epfl/sdp/mobile/state/StatefulGameScreen.kt
```diff
@@ -0,0 +1,93 @@
...
+import ch.epfl.sdp.mobile.ui.game.GameScreen
+import ch.epfl.sdp.mobile.ui.game.GameScreenState
+import ch.epfl.sdp.mobile.ui.game.Move
+
+/**
+ * Represents a [ChessBoardState]'s Move
+ * @property number The move's number in the game's history
+ * @property name The move's name in chess notation
+ */
+data class ChessMove(override val number: Int, override val name: String) : Move
```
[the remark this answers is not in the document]
matt989253: That would be quite impractical as it is used in `GameScreenTest`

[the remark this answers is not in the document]
matt989253: Duplicate

mobile/src/main/java/ch/epfl/sdp/mobile/ui/game/ChessBoardState.kt
alexandrepiveteau: The rationale was that the UI should not use the same types as the model, to separate concerns cleanly. In particular, the model enumeration may gain some behavior over time (to model moves for instance) which will be irrelevant to the UI layer. I think we do the same with profiles and people for example.

I agree that it looks a bit ridiculous right now. What would you say of revisiting this once we have a better idea of how the chess engine will be implemented, and how its `Rank` enumeration will look like ?

alexandrepiveteau (approved):
Thanks !

BadrTad (approved):

PR #100, opened 2022-03-15
Fix profile update deleting all existing fields

alexandrepiveteau:
When a profile was updated, all the fields (including the future list of following users !) were cleared. This PR includes a reproducer test as well the fix.


matt989253 (approved):

KurohanaJuri (approved):
thx for the fix

PR #109, opened 2022-03-16
Implement navigation to chess board from new game button

BadrTad:
This PR implements create a basic Play screen with a new game button only that navigate to chess board when clicked 

Misses:  
 - Hiding the bottom home bar when playing game
 - Documentation 
 - Tests
 - Solve merge confilcts with Matthieu's PR https://github.com/epfl-SDP/android/pull/97


mobile/src/main/java/ch/epfl/sdp/mobile/ui/home/HomeScaffold.kt
```diff
@@ -46,16 +46,19 @@ enum class HomeSection(
 fun HomeScaffold(
     section: HomeSection,
     onSectionChange: (HomeSection) -> Unit,
+    hideBar: Boolean,
```
alexandrepiveteau, line 49: Nitpick : I'd name this `hiddenBottomBar` instead, because :

- there can be multiple bars (top bar, bottom bar)
- I feel like `hide` may indicate an action, and `hidden` describes a state

mobile/src/main/java/ch/epfl/sdp/mobile/state/StatefulHome.kt
```diff
@@ -20,8 +20,11 @@ private const val SocialRoute = "social"
 /** The route associated to the settings tab. */
 private const val SettingsRoute = "settings"
 
-/** The route associated to the game tab. */
-private const val GameRoute = "game"
+/** The route associated to the play tab. */
+private const val PlayRoute = "play"
+
+/** The route associated to new game button */
+const val NewGameRoute = "new_game"
```
alexandrepiveteau, line 27: Maybe `GameRoute` is a better name, since the route doesn't describe that we created a new game but just on which screen we are.

alexandrepiveteau, line 27: I think this should be `private`. If you want to provide an action to navigate to a certain screen, you should use a callback in the signature of the stateful screen composable, so the `NavController` never leaves this `StatefulHome` composable actually.

mobile/src/main/java/ch/epfl/sdp/mobile/ui/Graphics.kt
```diff
@@ -29,51 +26,6 @@ val PawniesIcons.SectionSocial
 val PawniesIcons.SectionSettings
   get() = Icons.Default.Settings
 
-val PawniesIcons.Add
```
alexandrepiveteau, line 32: Removing these icons breaks the app, doesn't it ?

mobile/src/main/java/ch/epfl/sdp/mobile/ui/i18n/English.kt
```diff
@@ -16,16 +16,6 @@ object English : LocalizedStrings {
   override val authenticateErrorFailure = "Something went wrong when authenticating"
   override val authenticateOr = "or"
 
-  override val boardColorBlack = "black"
```
alexandrepiveteau, line 19: I think we need these strings for the game screen.

mobile/src/main/java/ch/epfl/sdp/mobile/ui/play/PlayScreenState.kt
```diff
@@ -0,0 +1,7 @@
+package ch.epfl.sdp.mobile.ui.play
+
+interface PlayScreenState {
+
+  val onNewGame: () -> Unit
```
alexandrepiveteau, line 5: Maybe name this `onNewGameClick`.

alexandrepiveteau (changes requested):
A good start, thanks !

alexandrepiveteau:
Regarding the padding issues you have with the new game button, you can use some `PaddingValues` returned by the `HomeScaffold` and pass them as an optional argument to your new stateful composable. Within the composable, you can just apply the provided padding with something like `Modifier.padding(contentPadding)` where needed (or even copy only some parts of the padding, like the bottom and end paddings).

```kotlin
@Composable
fun StatefulHome(
    user: AuthenticatedUser,
    controller: NavHostController = rememberNavController(),
) {
  val entry by controller.currentBackStackEntryAsState()
  val section = entry?.toSection() ?: HomeSection.Social
  HomeScaffold(
      section = section,
      onSectionChange = { controller.navigate(it.toRoute()) },
      modifier = modifier,
  ) { paddingValues -> // You can pass this as an optional argument to the sections.
    NavHost(
        navController = controller,
        startDestination = SocialRoute,
    ) {
      composable("Sample") {
        StatefulSection(
            user = user,
            modifier = Modifier.fillMaxSize(),
            contentPadding = paddingValues,
        )
      }
    }
  }
}

@Composable
fun StatefulSection(
    user: AuthenticatedUser,
    contentPadding: PaddingValues = PaddingValues(),
) {}
```

mobile/src/main/java/ch/epfl/sdp/mobile/ui/i18n/English.kt
BadrTad: Yes I rectified that

mobile/src/main/java/ch/epfl/sdp/mobile/ui/Graphics.kt
BadrTad: Yes I corrected it

mobile/src/main/java/ch/epfl/sdp/mobile/ui/home/HomeScaffold.kt
BadrTad: Agree

mobile/src/main/java/ch/epfl/sdp/mobile/state/StatefulHome.kt
BadrTad: Agree

BadrTad: So we mean there is no need to pass the `NavHostController` to the `StatefulPlayScreen `which then uses it to create  a callback for the `PlayScreen`

mobile/src/main/java/ch/epfl/sdp/mobile/ui/i18n/English.kt
```diff
@@ -16,16 +16,6 @@ object English : LocalizedStrings {
...
-  override val boardColorWhite = "white"
-  override val boardPieceKing = "king"
-  override val boardPieceQueen = "queen"
-  override val boardPieceRook = "rook"
-  override val boardPieceBishop = "bishop"
-  override val boardPieceKnight = "knight"
-  override val boardPiecePawn = "pawn"
-  override val boardPieceContentDescription = { color: String, rank: String -> "$color $rank" }
-
```
KurohanaJuri, line 28: This issue bring by @alexandrepiveteau is still not resolve

mobile/src/main/java/ch/epfl/sdp/mobile/state/StatefulPlayScreen.kt
```diff
@@ -0,0 +1,36 @@
...
+ * A stateful implementation of the PlayScreen
+ * @param controller Navigation controller to navigate to components
+ * @param modifier the [Modifier] for this composable.
+ */
+@Composable
+fun StatefulPlayScreen(
+    controller: NavHostController,
+    contentPadding: PaddingValues = PaddingValues(),
+    modifier: Modifier = Modifier,
+) {
```
KurohanaJuri, line 32: `Modifier` should be the first arg with default value

```suggestion
fun StatefulPlayScreen(
    controller: NavHostController,
    modifier: Modifier = Modifier,
    contentPadding: PaddingValues = PaddingValues(),
) {
```

mobile/src/main/java/ch/epfl/sdp/mobile/state/StatefulHome.kt
BadrTad: I don't see why would it be a bad thing to restrict the usage of the `NavController` in `StatefulHome` only. It could also be useful to pass it to `StatefulPlayScreen` for example.

mobile/src/main/java/ch/epfl/sdp/mobile/ui/play/PlayScreen.kt
```diff
@@ -0,0 +1,50 @@
...
+ * @param modifier the [Modifier] for this composable.
+ */
+@Composable
+fun NewGameButton(onNewGame: () -> Unit, modifier: Modifier = Modifier) {
+  val strings = LocalLocalizedStrings.current
+  Button(
+      onClick = onNewGame,
+      shape = CircleShape,
+      contentPadding = PaddingValues(horizontal = 24.dp, vertical = 16.dp),
+      modifier = modifier.padding(end = 16.dp, bottom = 16.dp)) {
```
alexandrepiveteau, line 45: (minor) Using `Modifier.padding(16.dp)` would work too here, and maybe be slightly easier to understand.

mobile/src/main/java/ch/epfl/sdp/mobile/ui/play/PlayScreen.kt
```diff
@@ -0,0 +1,50 @@
...
+    Icon(Icons.Default.Add, null)
```
alexandrepiveteau, line 46: You should add this icon to `PawniesIcons`.

mobile/src/main/java/ch/epfl/sdp/mobile/ui/play/PlayScreen.kt
```diff
@@ -0,0 +1,50 @@
...
+
+/**
+ * Composable that composes a new game button
+ * @param onNewGame callable upon button action
+ * @param modifier the [Modifier] for this composable.
+ */
+@Composable
+fun NewGameButton(onNewGame: () -> Unit, modifier: Modifier = Modifier) {
+  val strings = LocalLocalizedStrings.current
+  Button(
```
alexandrepiveteau, line 41: You may want to look into the `ExtendedFloatingActionButton` composable : https://developer.android.com/reference/kotlin/androidx/compose/material/package-summary#ExtendedFloatingActionButton(kotlin.Function0,kotlin.Function0,androidx.compose.ui.Modifier,kotlin.Function0,androidx.compose.foundation.interaction.MutableInteractionSource,androidx.compose.ui.graphics.Shape,androidx.compose.ui.graphics.Color,androidx.compose.ui.graphics.Color,androidx.compose.material.FloatingActionButtonElevation).

mobile/src/main/java/ch/epfl/sdp/mobile/state/StatefulHome.kt
alexandrepiveteau: It's really a matter of encapsulating how navigation works. The fact that we're using compose's `NavController` is really an implementation detail of the `StatefulHome`; some alternatives exist, such as Accompanist's https://github.com/google/accompanist/tree/main/navigation-animation for instance, and it'd be better if we only needed to change one Composable rather than all the `StatefulXXX` composables if we migrate to it for instance (especially if we want to have some proper screen transitions at some point).

mobile/src/androidTest/java/ch/epfl/sdp/mobile/test/state/StatefulHomeTest.kt
```diff
@@ -48,4 +48,29 @@ class StatefulHomeTest {
...
+    val api = AuthenticationFacade(emptyAuth(), emptyStore())
+    api.signUpWithEmail("email", "name", "password")
+    val user = api.currentUser.filterIsInstance<AuthenticatedUser>().first()
+    val strings = rule.setContentWithLocalizedStrings { StatefulHome(user) }
+    rule.onNodeWithText(strings.sectionPlay).performClick()
+    rule.onNodeWithText(strings.sectionPlay).assertIsSelected()
+    rule.onNodeWithText(strings.sectionSocial).assertIsNotSelected()
+    rule.onNodeWithText(strings.newGame).assertExists()
+    rule.onNodeWithText(strings.newGame).performClick()
+    rule.onNodeWithText(strings.newGame).assertDoesNotExist()
```
KurohanaJuri, line 74: @alexandrepiveteau Question, can we chain the action ? I read the doc, for me, it's possible to write something like this

```
rule.onNodeWithText(strings.newGame).assertExists()
     .performClick()
     .assertDoesNotExist()
```

I think it's less redundant and easier to read

alexandrepiveteau: I don't know at all if it behaves exactly the same, but I agree that it's less redundant and that it's worth trying it out !

mobile/src/main/java/ch/epfl/sdp/mobile/ui/play/PlayScreen.kt
```diff
@@ -0,0 +1,50 @@
...
+import androidx.compose.material.icons.Icons
+import androidx.compose.material.icons.filled.Add
+import androidx.compose.runtime.Composable
+import androidx.compose.ui.Alignment
+import androidx.compose.ui.Modifier
+import androidx.compose.ui.unit.dp
+import ch.epfl.sdp.mobile.state.LocalLocalizedStrings
+
+/**
+ * Composable that composes the PlayScreen [FIXME] Contains a new game button only, should be
```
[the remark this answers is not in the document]
BadrTad: That this is is not the final implementation in the sense that the `PlayScreen` requires other components like the history of matches.  It's rather a TODO than a FIXME.

mobile/src/androidTest/java/ch/epfl/sdp/mobile/test/ui/play/PlayScreenTest.kt
```diff
@@ -0,0 +1,23 @@
...
+
+import androidx.compose.ui.test.junit4.createComposeRule
+import androidx.compose.ui.test.onNodeWithText
+import ch.epfl.sdp.mobile.test.state.setContentWithLocalizedStrings
+import ch.epfl.sdp.mobile.ui.play.PlayScreen
+import ch.epfl.sdp.mobile.ui.play.PlayScreenState
+import org.junit.Rule
+import org.junit.Test
+
+open class TestPlayScreenState(override val onNewGameClick: () -> Unit) : PlayScreenState
```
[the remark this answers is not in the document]
BadrTad: Yes it's indeed sufficient for now, but we will definitely need both once `TestPlayScreenState` have more attributes like an authenticated user or matche history in order to perform other tests.

mobile/src/main/java/ch/epfl/sdp/mobile/state/StatefulHome.kt
BadrTad: Agree !

mobile/src/main/java/ch/epfl/sdp/mobile/state/StatefulPlayScreen.kt
```diff
@@ -0,0 +1,34 @@
...
+ * @param modifier the [Modifier] for this composable.
+ */
+@Composable
+fun StatefulPlayScreen(
+    navigateToGame: () -> Unit,
+    modifier: Modifier = Modifier,
+    contentPadding: PaddingValues = PaddingValues(),
+) {
+  /* TODO: Add any necessary args: authenticated user ... */
+  val state = remember { Player(newGame = navigateToGame) }
```
alexandrepiveteau, line 32:
```suggestion
  val state = remember(navigateToGame) { Player(newGame = navigateToGame) }
```

mobile/src/main/java/ch/epfl/sdp/mobile/ui/play/PlayScreen.kt
```diff
@@ -0,0 +1,52 @@
...
+ * @param modifier the [Modifier] for this composable.
+ */
+@Composable
+fun NewGameButton(onNewGame: () -> Unit, modifier: Modifier = Modifier) {
+  val strings = LocalLocalizedStrings.current
+  ExtendedFloatingActionButton(
+      text = { Text(strings.newGame) },
+      onClick = onNewGame,
+      shape = CircleShape,
+      modifier = modifier.padding(16.dp).size(169.dp, 56.dp),
```
alexandrepiveteau, line 49:
```suggestion
      modifier = modifier.padding(16.dp),
```

mobile/src/main/java/ch/epfl/sdp/mobile/ui/play/PlayScreen.kt
```diff
@@ -0,0 +1,52 @@
...
+@Composable
+fun PlayScreen(
+    state: PlayScreenState,
+    modifier: Modifier = Modifier,
+    contentPadding: PaddingValues = PaddingValues(),
+) {
+  Box(
+      contentAlignment = Alignment.BottomEnd,
+      modifier = modifier.fillMaxSize().padding(contentPadding)) {
+    NewGameButton(onNewGame = state.onNewGameClick)
```
alexandrepiveteau, line 33: Prefer aligning the individual Button component rather than all the children of the `Box`.

```suggestion
  Box(modifier = modifier.fillMaxSize().padding(contentPadding)) {
    NewGameButton(onNewGame = state.onNewGameClick, modifier = Modifier.align(Alignment.BottomEnd))
```

mobile/src/main/java/ch/epfl/sdp/mobile/state/StatefulPlayScreen.kt
```diff
@@ -0,0 +1,34 @@
...
+import androidx.compose.foundation.layout.PaddingValues
+import androidx.compose.runtime.Composable
+import androidx.compose.runtime.remember
+import androidx.compose.ui.Modifier
+import ch.epfl.sdp.mobile.ui.play.PlayScreen
+import ch.epfl.sdp.mobile.ui.play.PlayScreenState
+
+/**
+ * Player class that holds info about chess player
+ * @param newGame Callable action for newGame
```
KurohanaJuri, line 12: I don't understand what the action do, can you add more detail in the doc ?

mobile/src/main/java/ch/epfl/sdp/mobile/ui/play/PlayScreen.kt
```diff
@@ -0,0 +1,50 @@
...
+import androidx.compose.ui.Alignment
+import androidx.compose.ui.Modifier
+import androidx.compose.ui.unit.dp
+import ch.epfl.sdp.mobile.state.LocalLocalizedStrings
+import ch.epfl.sdp.mobile.ui.Add
+import ch.epfl.sdp.mobile.ui.PawniesColors
+import ch.epfl.sdp.mobile.ui.PawniesIcons
+
+/**
+ * Composable that composes the PlayScreen [TODO] Contains a new game button only, should be
```
KurohanaJuri, line 18: `[TODO]` ? What do you mean ?

mobile/src/androidTest/java/ch/epfl/sdp/mobile/test/state/StatefulHomeTest.kt
```diff
@@ -66,4 +66,27 @@ class StatefulHomeTest {
...
+  @Test
+  fun clickNewGame_inPlaySection_switchesToBoard() = runTest {
+    val facade = AuthenticationFacade(emptyAuth(), emptyStore())
+    facade.signUpWithEmail("email", "name", "password")
+    val user = facade.currentUser.filterIsInstance<AuthenticatedUser>().first()
+    val strings = rule.setContentWithLocalizedStrings { StatefulHome(user) }
+    rule.onNodeWithText(strings.sectionPlay).performClick()
+    rule.onNodeWithText(strings.sectionPlay).assertIsSelected()
+    rule.onNodeWithText(strings.sectionSocial).assertIsNotSelected()
+    rule.onNodeWithText(strings.newGame).assertExists().performClick().assertDoesNotExist()
```
KurohanaJuri, line 90: As you can see on Cirrus, some test fail. I search a bit in your branch and the report says that the Social API is missing.

mobile/src/main/java/ch/epfl/sdp/mobile/state/StatefulPlayScreen.kt
BadrTad: Ofc

mobile/src/main/java/ch/epfl/sdp/mobile/ui/play/PlayScreen.kt
BadrTad: I agree it's not really clear clear. What I meant is :
`[TODO] Contains a new game button only, should be expanded to include history of matches`

mobile/src/androidTest/java/ch/epfl/sdp/mobile/test/state/StatefulHomeTest.kt
BadrTad: Thanks it's fixed now

KurohanaJuri (approved):
Thx for the fix !

PR #113, opened 2022-03-17
Remove generic `Identifier` from piece and make `Piece` generic

alexandrepiveteau:


matt989253 (approved):
LGTM, makes it quite a bit clearer!

BadrTad (approved):
Code looks cleaner indeed !

PR #114, opened 2022-03-17
Center the email and name properly in the profile section

alexandrepiveteau:
As noted by @KurohanaJuri, the email isn't properly centred in the profile screen. This PR fixes that issue by specifying the right column alignment.


KurohanaJuri (approved):
Thx for the fix

matt989253 (approved):

PR #134, opened 2022-03-19
Add chess movement rules

alexandrepiveteau:
This PR adds support for standard moves for all the pieces, except those which are "history-related" (aka. castling and en-passant). Additionally, the `ChessBoard` composable now displays dots on all the squares on which a piece may be dropped.

To make moves agnostic of the color of the current player, the board is normalised before computing which moves are available. When normalised, players don't see piece colors anymore but only if pieces belong to themselves or to their adversary, and a vertical symmetry is applied if the current player is black (so both players see the board as if they were playing as white). The `Color` enumeration is extended to provide support for board normalisation.

The `Rank` enumeration is extended to provide access to `Moves` specific to the `Rank`. Moves are computed on a normalised board.


mobile/src/main/java/ch/epfl/sdp/mobile/application/chess/Color.kt
```diff
@@ -1,11 +1,53 @@
...
+   * Normalizes the given [Position].
+   *
+   * @param position the [Position] to normalize.
+   * @return the normalized position.
+   */
+  fun normalize(position: Position): Position = normalizeToWhitePosition(position)
+
+  /** Returns the [Color] of the adversary. */
+  fun other(): Color = opposite()
 }
```
matt989253, line 53: Since both Black and White get normalized, maybe the method `normalizeToWhitePosition` should actually be called `normalizePosition`, same thing with `normalizeDelta`?

mobile/src/main/java/ch/epfl/sdp/mobile/application/chess/Game.kt
```diff
@@ -1,14 +1,63 @@
...
+        }
+
+        // Populate the pieces.
+        populateSide(0, Black)
+        populateSide(7, White)
+
+        // Populate the pawns.
+        repeat(Board.Size) { column ->
+          set(Position(column, 1), Piece(Black, Pawn, id++))
+          set(Position(column, 6), Piece(White, Pawn, id++))
```
matt989253, line 58: Should the `inc()` method not be called instead of `id++`?

mobile/src/main/java/ch/epfl/sdp/mobile/application/chess/rules/Moves.kt
```diff
@@ -0,0 +1,144 @@
...
+ * @param from the original [Position].
+ * @param row the row on which the double-up move is allowed.
+ * @param includeAdversary true if the [Moves] should include stepping on and eating adversaries.
+ */
+fun Board<Piece<Role>>.doubleUp(
+    from: Position,
+    row: Int = 6,
+    includeAdversary: Boolean = false,
+): Moves = sequence {
+  if (from.y != row) return@sequence // Not on the right row.
```
matt989253, line 49: From a chess rule perspective, this feels wrong... but it is not...

mobile/src/main/java/ch/epfl/sdp/mobile/application/chess/Game.kt
alexandrepiveteau: The `inc()` method is an `operator fun` which is called when the increment operator is applied to `id` (it's essentially the reason why I defined `inc` on `PersistentPieceIdentifier`). So here `id++` is translated by the compiler to `id = id.inc()`.

mobile/src/main/java/ch/epfl/sdp/mobile/application/chess/Color.kt
alexandrepiveteau: Fixed in affda0427dcb613a4afb6ef6c53c787c5213f072.

matt989253 (approved):
Thanks for the doc, helped a lot in reviewing! Looks good to me now.

mobile/src/main/java/ch/epfl/sdp/mobile/application/chess/Game.kt
```diff
@@ -1,14 +1,71 @@
...
+   * does not have a [Piece] or it's not this player's turn, the resulting sequence might be empty.
+   *
+   * @param position the position for which the available [Action]s are queried.
+   * @return the [Sequence] of available actions.
+   */
+  fun actions(position: Position): Sequence<Action>
 }
+
+/** Creates a new [Game], with the standard starting positions for both players. */
+fun emptyGame(): Game =
```
KurohanaJuri, line 35: [Suggestion] Can we change it to `initializeGame`. `emtpyGame`isn't clear enough for me

alexandrepiveteau: Your concern is valid, however `initializeGame` would not really follow standard Kotlin naming conventions (whereas `emptyGame` would, but I agree that it should not result in an non-empty board being created). Would a method like `Game.create()` (on the `Game` `companion object` sound good to you ?

KurohanaJuri: It's fine for me !

alexandrepiveteau: Done in 65a6e465349af7486bd98dc271922d4956524656.

matt989253 (approved):

mobile/src/main/java/ch/epfl/sdp/mobile/application/chess/implementation/PersistentBoard.kt
```diff
@@ -0,0 +1,68 @@
...
+ * An implementation of a [Board] which is backed on a persistent collection.
+ *
+ * @param Piece the type of the pieces contained in this board.
+ */
+data class PersistentBoard<Piece : Any>
+constructor(
+    private val cells: PersistentMap<Position, Piece>,
+) : Board<Piece> {
+
+  override fun get(position: Position): Piece? = cells[position]
```
BadrTad, line 56: [Question] Why do you sometimes use get or set  as `operator` and aometimes you don't.  For example here

mobile/src/main/java/ch/epfl/sdp/mobile/application/chess/Color.kt
```diff
@@ -1,11 +1,53 @@
...
+enum class Color(
+    private val normalizeDelta: Delta.() -> Delta,
+    private val normalizePosition: Position.() -> Position,
+    private val opposite: () -> Color,
+) {
 
   /** The black color. */
-  Black,
+  Black(
+      normalizeDelta = { Delta(x = x, y = -y) },
```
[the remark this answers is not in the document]
alexandrepiveteau: The `x` and `-y` come from the `Delta.()` receiver (since `normalizeDelta` is a `Delta.() -> Delta`. When normalised, a `Delta` for the black player should have a vertical symmetry applied, hence why the `y` coordinate is inverted.

mobile/src/main/java/ch/epfl/sdp/mobile/application/chess/Color.kt
```diff
@@ -1,11 +1,53 @@
...
+            y = Board.Size - y - 1, // A simple vertical symmetry.
+        )
+      },
+      opposite = { White },
+  ),
 
   /** The white color. Starts the game. */
-  White,
+  White(
+      normalizeDelta = { this },
```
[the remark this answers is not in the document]
alexandrepiveteau: `this` references the `Delta.()` argument in the `normalizeDelta` signature (`Delta.() -> Delta`).

mobile/src/main/java/ch/epfl/sdp/mobile/application/chess/implementation/PersistentBoard.kt
alexandrepiveteau: For no specific reason other than Kotlin's lenience when it comes to enforcing that functions are marked as `operator` 😄 

I'll update the PR to add them systematically.

Fouad-sys (approved):

mobile/src/main/java/ch/epfl/sdp/mobile/application/chess/implementation/PersistentBoard.kt
alexandrepiveteau: Welp, the PR got merged before I was able to push my commit.

mobile/src/main/java/ch/epfl/sdp/mobile/application/chess/Board.kt
```diff
@@ -1,7 +1,11 @@
...
-interface Board {
+/**
+ * A [Board] contains a bunch of [Piece] positioned in a certain way.
+ *
+ * @param Piece the type of the pieces which are present in a board.
+ */
+interface Board<Piece : Any> {
 
   /**
    * Returns the [Piece] at a certain [Position] in the board, or null if no [Piece] is localed
```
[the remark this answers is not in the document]
alexandrepiveteau: I think I meant located

mobile/src/main/java/ch/epfl/sdp/mobile/application/chess/Board.kt
```diff
@@ -1,7 +1,11 @@
 package ch.epfl.sdp.mobile.application.chess
 
-/** A [Board] contains a bunch of [Piece] positioned in a certain way. */
-interface Board {
+/**
+ * A [Board] contains a bunch of [Piece] positioned in a certain way.
```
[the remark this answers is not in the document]
alexandrepiveteau: Adding an 's' after the type name is something I sometimes do, so both approaches sound good to me. I think we've been a bit lenient with documentation in the project until now, so maybe it's worth discussing with the team if / how we want to improve docs

PR #145, opened 2022-03-21
Add AR pawn model

KurohanaJuri:
Create a new tab in the navigation that show the 2 AR pawn

I don't know how we can test this, any suggestion is welcome


gradle/libs.versions.toml
```diff
@@ -45,6 +45,8 @@ mockk-mockk = { module = "io.mockk:mockk", version.ref = "mockk" }
 mockk-android = { module = "io.mockk:mockk-android", version.ref = "mockk" }
 mockk-agent-jvm = { module = "io.mockk:mockk-agent-jvm", version.ref = "mockk" }
 
+arscenveview = "io.github.sceneview:arsceneview:0.2.0"
```
alexandrepiveteau, line 48: Typo : I think this dependency should be named `arsceneview`

mobile/build.gradle.kts
```diff
@@ -38,7 +38,10 @@ android {
 
   composeOptions { kotlinCompilerExtensionVersion = libs.versions.compose.get() }
   packagingOptions { resources.excludes.add("META-INF/*") }
-  buildFeatures { compose = true }
+  buildFeatures {
+    viewBinding = true
```
alexandrepiveteau, line 42: Question : do we really need `viewBinding` ?

mobile/build.gradle.kts
```diff
@@ -53,7 +56,6 @@ tasks.withType<KotlinCompile>().configureEach {
 }
 
 dependencies {
-
```
alexandrepiveteau, line 56: Nitpick : I don't think this formatting change is needed

mobile/build.gradle.kts
```diff
@@ -84,6 +86,8 @@ dependencies {
 
   // Jetpack Compose
   implementation(libs.bundles.compose.android)
+
+  implementation("io.github.sceneview:arsceneview:0.2.0")
```
alexandrepiveteau, line 90: Suggestion : you should refer to `libs.arsceneview` once you've renamed the dependency in the Gradle version catalog

mobile/src/main/AndroidManifest.xml
```diff
@@ -24,6 +24,12 @@
             </intent-filter>
         </activity-alias>
 
+        <!-- "AR Optional" app, contains non-AR features that can be used when
+        "Google Play Services for AR" (ARCore) is not available. -->
+        <meta-data
+            android:name="com.google.ar.core"
+            android:value="optional" />
+
```
alexandrepiveteau, line 32: Question : is this really needed ? The merged Manifest seems to indicate that it is already included, since SceneView already declares this `meta-data` tag (https://github.com/SceneView/sceneview-android/blob/main/arsceneview/src/main/AndroidManifest.xml#L8-L14).

mobile/build.gradle.kts
KurohanaJuri: No, it was something that I forget to remove

mobile/src/main/AndroidManifest.xml
KurohanaJuri: You're right, I will remove it

mobile/build.gradle.kts
KurohanaJuri: Solve in commit https://github.com/epfl-SDP/android/pull/145/commits/8741bccf8827392c7f1bc51eb55343ed6f3506bc

barmettlerl (approved):
Looks good for a prototype

PR #150, opened 2022-03-23
Follow/Unfollow Button Animation

Fouad-sys:


mobile/src/main/java/ch/epfl/sdp/mobile/ui/social/FollowButton.kt
```diff
@@ -0,0 +1,87 @@
...
+/**
+ * An implementation of [ButtonColors] used to change the color of a button according to whether it
+ * is selected.
+ *
+ * @param selected a [State] of [Boolean] indicating whether the button is selected.
+ */
+private class FollowButtonColors(selected: State<Boolean>) : ButtonColors {
+
+  /** A [State] of [Boolean] indicating whether the button is selected. */
+  private val selected by selected
```
KurohanaJuri, line 32: Question : How can you assign a variable with the name ?

Fouad-sys: Not sure I completely understood but the other selected name refers to the class parameter

alexandrepiveteau (approved):
LGTM, thanks !

mobile/src/main/java/ch/epfl/sdp/mobile/ui/social/FollowButton.kt
alexandrepiveteau: Indeed, this is a property delegate for the class parameter `selected: State<Boolean>`.

matt989253 (approved):
Nice, thanks! LGTM

PR #151, opened 2022-03-23
Add check detection support to the game engine

alexandrepiveteau:
Changes :

- `BoardHistory` was removed and `BoardSequence` introduced to replace it. This is a breaking change because the `MovesTest` file was creating sequences of boards in reverse order. The tests have been updated to reflect the ordering change.
- `Moves` was made generic.
- `Board` now implements `Iterable<Pair<Position, Piece>>` so it can be iterated on.  
- `NextStep` has some new subclasses for checkmate and stalemate.


mobile/src/main/java/ch/epfl/sdp/mobile/application/chess/implementation/PersistentGame.kt
```diff
@@ -14,64 +14,109 @@ import ch.epfl.sdp.mobile.application.chess.rules.Role
...
   }
 }
+
+/**
+ * Computes all the [Moves] which can be performed by a player of the given [color].
+ *
+ * @param color the [Color] of the player for whom all the moves are computed.
+ */
+private fun BoardSequence<Piece<Color>>.allMoves(color: Color): Moves<Piece<Color>> {
+  return first().asSequence().flatMap { (pos, _) -> allMovesAtPosition(color, pos) }
```
BadrTad, line 102: [Question] Why are you using `first()` and not `last()` for `BoardSequence` isn't the latter supposed to be the most updated one ?

BadrTad (commented):
Great work. Thanks in advance for your clarifications.

mobile/src/main/java/ch/epfl/sdp/mobile/application/chess/implementation/NormalizedBoard.kt
```diff
@@ -62,4 +66,8 @@ class DenormalizedBoardDecorator(
     val result = board.set(player.normalize(position), piece?.denormalize())
     return DenormalizedBoardDecorator(player, result)
   }
+
+  override fun iterator(): Iterator<Pair<Position, Piece<Color>>> {
+    return board.asSequence().map { (pos, piece) -> pos to piece.normalize() }.iterator()
```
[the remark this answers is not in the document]
alexandrepiveteau: The normalisation / denormalisation processes are actually identical since it only consists in a vertical symmetry for the black player. However I'll see if / how I can rename things to make them more consistent with the name of the decorator.

mobile/src/main/java/ch/epfl/sdp/mobile/application/chess/implementation/PersistentGame.kt
alexandrepiveteau: The `Sequence` provides access to the most recent board state first. This way, since `Sequence` is a lazy collection, moves that only need to access the most recent states can iterate on the first values of the collection.

matt989253 (approved):
LGTM

Fouad-sys (approved):
Looks great to me, thank you!

Issue #153, opened 2022-03-25
Refactor online chess files


matt989253:
Notably:
- [x] StoreDocuments are bloated with serializing functions that should have their own files
- [x] Document `Match` companion object
- [x] Add a dedicated Robot to test the `StatefulGameScreen`
- [x] Add test for when the CompositionLocal is missing.
- [x] Increase code coverage, especially for the serialization functions 
- [x] Find a better name for `ChessFacade`'s `fetchMatchesForUser` (e.g. `matches`, since it fetches a Flow)
- [x] Move `ChessFacade`s `updateMatch` to be a function on ``Match`

PR #176, opened 2022-03-25
Add a robot for the `ChessBoard` composable

alexandrepiveteau:
This is part of the refactoring required of #153. This pull request adds a `ChessBoardRobot`, and updates some existing chessboard tests to use it.

### DIFF
--- a/mobile/src/androidTest/java/ch/epfl/sdp/mobile/test/ui/Layout.kt
+++ b/mobile/src/androidTest/java/ch/epfl/sdp/mobile/test/ui/Layout.kt
@@ -1,5 +1,7 @@
 package ch.epfl.sdp.mobile.test.ui
 
+import androidx.compose.ui.geometry.Offset
+import androidx.compose.ui.geometry.Rect
 import androidx.compose.ui.semantics.SemanticsNode
 import androidx.compose.ui.unit.DpOffset
 import androidx.compose.ui.unit.DpRect
@@ -25,5 +27,16 @@ fun SemanticsNode.getBoundsInRoot(): DpRect =
  * @return true if the offset is contained.
  */
 operator fun DpRect.contains(offset: DpOffset): Boolean {
+  return offset.x in left..right && offset.y in top..bottom
+}
+
+/**
+ * Returns true iff the [Offset] is contained within this [Rect] bounds.
+ *
+ * @receiver the bounds of the [Rect].
+ * @param offset the offset that we're checking some bounds for.
+ * @return true if the offset is contained.
+ */
+operator fun Rect.contains(offset: Offset): Boolean {
   return offset.x in left..right && offset.y in top..bottom
 }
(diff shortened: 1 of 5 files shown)


mobile/src/androidTest/java/ch/epfl/sdp/mobile/test/ui/Layout.kt
```diff
@@ -27,3 +29,14 @@ fun SemanticsNode.getBoundsInRoot(): DpRect =
...
+
+/**
+ * Returns true iff the [Offset] is contained within this [Rect] bounds.
+ *
+ * @receiver the bounds of the [Rect].
+ * @param offset the offset that we're checking some bounds for.
+ * @return true if the offset is contained.
+ */
+operator fun Rect.contains(offset: Offset): Boolean {
+  return offset.x in left..right && offset.y in top..bottom
```
barmettlerl, line 41: What is the `..`? Never saw this before

mobile/src/androidTest/java/ch/epfl/sdp/mobile/test/ui/game/ChessBoardRobot.kt
```diff
@@ -0,0 +1,194 @@
...
+   * Moves the pointer with the given id to this cell.
+   *
+   * @param x the x coordinate of the target cell.
+   * @param y the y coordinate of the target cell.
+   * @param pointerId the identifier of the pointer.
+   */
+  fun moveTo(x: Int, y: Int, pointerId: Int = 0)
+
+  /**
+   * Moves the pointer with the given id by a relative amount of cells.
```
barmettlerl, line 41: Question: what do you mean by pointer?

barmettlerl (approved):

mobile/src/androidTest/java/ch/epfl/sdp/mobile/test/ui/Layout.kt
alexandrepiveteau: This is the `rangeTo` operator ([reference](https://kotlinlang.org/api/latest/jvm/stdlib/kotlin.ranges/range-to.html)).

mobile/src/androidTest/java/ch/epfl/sdp/mobile/test/ui/game/ChessBoardRobot.kt
alexandrepiveteau: On multi-touch devices, gestures are composed of multiple pointers (usually the fingers of the user). This terminology is used by Android gesture system (see the documentation for [`MotionEvent`](https://developer.android.com/reference/android/view/MotionEvent) for instance). Compose uses the same terminology.

The default pointer has value 0. If you look into Jetpack Compose's `SemanticsNodeInteraction.performTouchInput { }` method, you'll see that gestures are performed with this pointer by default, but allow the user to specify another pointer identifier. The `ChessBoardRobot` essentially allows the same behaviour, so we can test multi-touch gestures too.

Fouad-sys (approved):
LGTM!

mobile/src/androidTest/java/ch/epfl/sdp/mobile/test/ui/Layout.kt
Fouad-sys: Was going to ask the same question, thanks for clearing that up!

PR #179, opened 2022-03-27
Add a pawn on the board in the correct position

### LINKED ISSUE #161: AR: Put one chess figure at the right spot on the chess board

KurohanaJuri:
This PR adds a board model and a pawn on a specified position. I change the model placement mode, the model will be displayed on the taped position. This mode is more user-friendly and remove some strange behaviours. 

Some values are hardcoded : 
- Chessboard border : This value cannot be computed, we find it by guessing the value
- The position : It's a temporary `ChessBoardState.Position` value to simulate what we have in the chessboard state

This PR contains also some code used to debug that I don't remove deliberately. It's allow use to display 3D axes, which is kind useful when you don't understand what happen.

Closed #161


mobile/src/main/java/ch/epfl/sdp/mobile/ui/ar/ArScreen.kt
```diff
@@ -32,42 +60,107 @@ fun ArScreen(modifier: Modifier = Modifier) {
...
-   * @param modelNode the model that we want to add
-   * @param view Where the model will be added
-   */
-  fun addNode(modelNode: ArModelNode?, view: ArSceneView) {
-    val node = modelNode ?: return
-    if (node !in view.children) {
-      view.addChild(node)
-    }
+  // Keep the screen only for this composable
+  DisposableEffect(Unit) {
```
alexandrepiveteau, line 72: It's very unlikely (probably not even possible) that the `LocalView` changes but you should run the `DisposableEffect` again if it does.

```suggestion
  DisposableEffect(view) {
```

mobile/src/main/java/ch/epfl/sdp/mobile/ui/ar/ArScreen.kt
```diff
@@ -32,42 +60,107 @@ fun ArScreen(modifier: Modifier = Modifier) {
...
+        fun anchorOrMoveBoard(anchor: Anchor) {
+          // Add only one instance of the node
+          if (!view.children.contains(board!!)) {
+            view.addChild(board!!)
+          }
+          board?.anchor = anchor
+        }
+
+        // Place the board on the taped position
+        view.onTouchAr = { hitResult, _ -> anchorOrMoveBoard(hitResult.createAnchor()) }
```
alexandrepiveteau, line 164: Question : what happens if the callback is called before `board` is loaded ? Could the `board` be null and the app crash ?

mobile/src/main/java/ch/epfl/sdp/mobile/ui/ar/ArScreen.kt
```diff
@@ -32,42 +60,107 @@ fun ArScreen(modifier: Modifier = Modifier) {
...
+          }
+        }
+
+        /**
+         * TODO : With the [ChessBoardState], iterate over the list of pieces, and do the same that
+         * [pawn1]
+         */
+        if (pawn1 != null) {
+          addPiece(pawn1!!)
+        }
```
alexandrepiveteau, line 143: Suggestion : to avoid force-unwrapping, you could use a `let` expression :

```kotlin
pawn1?.let { addPiece(it) }
```

mobile/src/main/java/ch/epfl/sdp/mobile/ui/ar/ArScreen.kt
```diff
@@ -32,42 +60,107 @@ fun ArScreen(modifier: Modifier = Modifier) {
...
+            it.scale(2f)
+            board?.addChild(it)
+            it.placementPosition = Position(z = 3f, y = arBoard!!.boardHeight)
+          }
+          green?.let {
+            it.scale(2f)
+            board?.addChild(it)
+            it.placementPosition = Position(y = 3f + arBoard!!.boardHeight)
+          }
+        }
```
alexandrepiveteau, line 126: Question : is there a scenario where `white`, `red`, `blue` or `green` might be non-null but `DisplayAxes` is false ?

alexandrepiveteau (commented):
Looking great ! The new placement is much better, and it's great seeing things shape up well on AR.

In general, I feel like there are many force-unwrapping (mostly due to the fact that many of the `val` use property delegates), and it feels like we're loosing a bit of Kotlin's compile-time null-safety.

Imho the easiest way to solve that would be to assign the value to a separate `val`, and use smart-casting (something along these lines) :

```kotlin
val board by remember { mutableStateOf<ArSomeNullableType?>(null) }

fun something() {
  val currentBoard = board ?: return
  // Now we can use currentBoard safely
  currentBoard.something()
}
```

What are your thoughts on this ?

mobile/src/main/java/ch/epfl/sdp/mobile/ui/ar/ArScreen.kt
KurohanaJuri: Indeed, if the board is null, the app crash at line 157. We have a null safety on the `board`.

mobile/src/main/java/ch/epfl/sdp/mobile/ui/ar/ArScreen.kt
KurohanaJuri: Non because we only load the model if `DisplayAxes` is true (line 93-98).

KurohanaJuri:
> Imho the easiest way to solve that would be to assign the value to a separate `val`, and use smart-casting (something along these lines) :
> 
> ```kotlin
> val board by remember { mutableStateOf<ArSomeNullableType?>(null) }
> 
> fun something() {
>   val currentBoard = board ?: return
>   // Now we can use currentBoard safely
>   currentBoard.something()
> }
> ```

I don't understand your example. Why do you call `something()` again, it's a recursion that never stop, right ?

@alexandrepiveteau

alexandrepiveteau:
The method `ArSomeNullableType.something()` has nothing to do with the top-level `something()` in this case, I just happened to name both the method and the function similarly. Maybe this is less confusing ?

```kotlin
val board by remember { mutableStateOf<ArSomeNullableType?>(null) }

fun something() {
  val currentBoard = board ?: return
  // Now we can use currentBoard safely
  currentBoard.somethingElse()
}
```

The goal of this example is just to show how using a new `val` helps us avoid force-unwrapping with `!!`, because the compiler can now know for sure a `val` is not null (which it can't do if the `val` is a delegate !).

(as a side-note, this would not create a recursion, since `currentBoard.something()` will never resolve to the top-level `fun something()`, but only on a function named something available on a `ArSomeNullableType`.

mobile/src/androidTest/java/ch/epfl/sdp/mobile/test/ui/ar/ArScreenKtTest.kt
```diff
@@ -0,0 +1,44 @@
...
+
+class ArScreenKtTest {
+
+  @get:Rule val permissionRule: GrantPermissionRule = grant(Manifest.permission.CAMERA)
+  @get:Rule val rule = createAndroidComposeRule<HomeActivity>()
+
+  @Test
+  fun check_thatComposableHasTheTag() = withCanceledIntents {
+    rule.setContent { ArScreen() }
+    rule.onNode(hasAnyDescendant(hasTestTag("AR"))).assertExists()
```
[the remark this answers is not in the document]
KurohanaJuri: Good idea, I changed this in the commit https://github.com/epfl-SDP/android/pull/179/commits/27f46f609b1c697a385475a2e5388f158a0a449a

mobile/src/main/java/ch/epfl/sdp/mobile/ui/ar/ArScreen.kt
KurohanaJuri: Fixed in https://github.com/epfl-SDP/android/pull/179/commits/198dc6e73229ae54fbf8dffde73c55071fb8ca03

matt989253 (dismissed):

alexandrepiveteau (dismissed):
LGTM, thanks !

alexandrepiveteau (approved):

mobile/src/main/java/ch/epfl/sdp/mobile/ui/ar/ArBoard.kt
```diff
@@ -0,0 +1,30 @@
...
+  private val cellSize = (boardHalfSize - boardBorderSize) / 4
+  private val cellCenter = cellSize / 2
+
+  /** For the given [ChessBoardState.Position] and transform it into AR board [Position] */
+  fun toArPosition(position: ChessBoardState.Position): Position {
+
+    fun transform(value: Int): Float {
+      return -boardHalfSize + boardBorderSize + value * cellSize + cellCenter
+    }
+    return Position(y = boardHeight, x = transform(position.x), z = transform(position.y))
```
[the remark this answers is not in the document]
KurohanaJuri: You're right, it looks weird. I change it in the commit https://github.com/epfl-SDP/android/pull/179/commits/040699df8fd274b68267fa18833d5dd40899b6c6

matt989253 (approved):

PR #181, opened 2022-03-27
Display the check status on the chessboard

alexandrepiveteau:
This PR adds a solid color indicator on the background the king pieces if the player is currently in check.

### DIFF
--- a/mobile/src/main/java/ch/epfl/sdp/mobile/state/StatefulGameScreen.kt
+++ b/mobile/src/main/java/ch/epfl/sdp/mobile/state/StatefulGameScreen.kt
@@ -74,6 +74,18 @@ class SnapshotChessBoardState(
   override var selectedPosition by mutableStateOf<ChessBoardState.Position?>(null)
     private set
 
+  override val checkPosition: ChessBoardState.Position?
+    get() {
+      val nextStep = match.game.nextStep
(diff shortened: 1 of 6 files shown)


mobile/src/main/java/ch/epfl/sdp/mobile/state/StatefulGameScreen.kt
```diff
@@ -74,6 +74,18 @@ class SnapshotChessBoardState(
   override var selectedPosition by mutableStateOf<ChessBoardState.Position?>(null)
     private set
 
+  override val checkPosition: ChessBoardState.Position?
+    get() {
+      val nextStep = match.game.nextStep
+      if (nextStep !is NextStep.MovePiece || !nextStep.inCheck) return null
```
matt989253, line 80: Question: Why is the `!is NextStep.MovePiece`  necessary?

matt989253 (approved):

mobile/src/main/java/ch/epfl/sdp/mobile/state/StatefulGameScreen.kt
alexandrepiveteau: If the game's `nextStep` is a `Stalemate` for instance, we don't want to highlight the king but display some other UI instead.

Fouad-sys (approved):
Looks great! Made me learn new Kotlin syntax too.

PR #182, opened 2022-03-27
Add an overload to specify the `uid` in `buildAuth { ... }` DSL

alexandrepiveteau:


mobile/src/androidTest/java/ch/epfl/sdp/mobile/test/infrastructure/persistence/auth/fake/FakeAuth.kt
```diff
@@ -46,32 +46,37 @@ class FakeAuth : Auth, AuthBuilder {
...
    * @param email the email address to use.
    * @param password the password of the added user.
+   * @param uid the unique identifier for this newly added user.
    *
    * @return the result of adding the new user.
    */
   private fun addUser(
       email: String,
       password: String,
+      uid: String = UUID.randomUUID().toString(),
```
barmettlerl, line 65: Question: Not sure if Random is better than default ("0") or so.

barmettlerl (approved):

mobile/src/androidTest/java/ch/epfl/sdp/mobile/test/infrastructure/persistence/auth/fake/FakeAuth.kt
alexandrepiveteau: I think it is, because UUID v4 has sufficient entropy for our practical needs, and it doesn't require checking if the "automatically-incremented" nextUid is already associated with another account when creating a new user.

matt989253 (approved):

PR #191, opened 2022-03-29
Add an `ExpandableFloatingActionButton` composable

alexandrepiveteau:
This PR adds an `ExpandableFloatingActionButton` composable. This composable represents a button that may be expanded to display some choices. The follows the metrics and guidelines described on Figma.

## Figma prototype

NewGameButton

## Implementation (using the usage sample code)

---

## Usage sample

> This sample is missing some dividers between the different cells, but they can easily be added in the `expandedContent` column.

```kotlin
@Preview
@Composable
fun Preview() = PawniesTheme {
  val state = rememberExpandableFloatingActionButtonState()

  LaunchedEffect(state) {
    while (true) {
      delay(3000)
      state.expanded = !state.expanded
    }
  }

  Box(
      Modifier.fillMaxSize().background(MaterialTheme.colors.background),
      Alignment.BottomEnd,
  ) {
    ExpandableFloatingActionButton(
        state = state,
        expandedContent = {
          ExpandableFloatingActionButtonItem(
              onClick = {},
              icon = { Icon(PawniesIcons.Search, null) },
              text = { Text("Local".uppercase()) },
              modifier = Modifier.fillMaxWidth(),
          )
          ExpandableFloatingActionButtonItem(
              onClick = {},
              icon = { Icon(PawniesIcons.SectionSettings, null) },
              text = { Text("Remote".uppercase()) },
              modifier = Modifier.fillMaxWidth(),
          )
        },
        collapsedContent = {
          Icon(PawniesIcons.Add, null)
          Spacer(Modifier.width(16.dp))
          Text("New game".uppercase())
        },
        modifier = Modifier.padding(16.dp),
    )
  }
}
```

### DIFF
--- a/mobile/src/androidTest/java/ch/epfl/sdp/mobile/test/ui/play/ExpandableFloatingActionButtonTest.kt
+++ b/mobile/src/androidTest/java/ch/epfl/sdp/mobile/test/ui/play/ExpandableFloatingActionButtonTest.kt
@@ -9,7 +9,7 @@ import androidx.compose.ui.test.junit4.createComposeRule
 import androidx.compose.ui.test.onNodeWithText
 import androidx.compose.ui.test.performClick
 import ch.epfl.sdp.mobile.ui.play.ExpandableFloatingActionButton
-import ch.epfl.sdp.mobile.ui.play.ExpandableFloatingActionButtonState
+import ch.epfl.sdp.mobile.ui.play.ExpandableFloatingActionButtonItem
 import com.google.common.truth.Truth.assertThat
 import org.junit.Rule
 import org.junit.Test
@@ -18,24 +18,29 @@ class ExpandableFloatingActionButtonTest {
 
   @get:Rule val rule = createComposeRule()
 
-  private class SnapshotExpandableFloatingActionButtonState(
-      initiallyExpanded: Boolean = false,
-  ) : ExpandableFloatingActionButtonState {
-    override var expanded by mutableStateOf(initiallyExpanded)
-  }
-
   @Test
   fun clickingButton_expandsIt() {
-    val state = SnapshotExpandableFloatingActionButtonState()
     rule.setContent {
       ExpandableFloatingActionButton(
-          state = state,
           expandedContent = { Text("Expanded") },
           collapsedContent = { Text("Collapsed") },
       )
     }
     rule.onNodeWithText("Collapsed").performClick()
     rule.onNodeWithText("Expanded").assertIsDisplayed()
-    assertThat(state.expanded).isTrue()
+  }
+
+  @Test
+  fun item_detectsClicks() {
+    var clicked by mutableStateOf(false)
+    rule.setContent {
+      ExpandableFloatingActionButtonItem(
+          onClick = { clicked = true },
+          icon = {},
+          text = { Text("Item") },
+      )
+    }
+    rule.onNodeWithText("Item").performClick()
+    assertThat(clicked).isTrue()
   }
 }
--- a/mobile/src/androidTest/java/ch/epfl/sdp/mobile/test/ui/play/ExpandableFloatingActionButtonTest.kt
+++ b/mobile/src/androidTest/java/ch/epfl/sdp/mobile/test/ui/play/ExpandableFloatingActionButtonTest.kt
@@ -1,0 +1,41 @@
+package ch.epfl.sdp.mobile.test.ui.play
+
+import androidx.compose.material.Text
+import androidx.compose.runtime.getValue
+import androidx.compose.runtime.mutableStateOf
+import androidx.compose.runtime.setValue
+import androidx.compose.ui.test.assertIsDisplayed
+import androidx.compose.ui.test.junit4.createComposeRule
+import androidx.compose.ui.test.onNodeWithText
+import androidx.compose.ui.test.performClick
+import ch.epfl.sdp.mobile.ui.play.ExpandableFloatingActionButton
+import ch.epfl.sdp.mobile.ui.play.ExpandableFloatingActionButtonState
+import com.google.common.truth.Truth.assertThat
+import org.junit.Rule
+import org.junit.Test
+
+class ExpandableFloatingActionButtonTest {
+
+  @get:Rule val rule = createComposeRule()
+
+  private class SnapshotExpandableFloatingActionButtonState(
+      initiallyExpanded: Boolean = false,
+  ) : ExpandableFloatingActionButtonState {
+    override var expanded by mutableStateOf(initiallyExpanded)
+  }
+
+  @Test
+  fun clickingButton_expandsIt() {
+    val state = SnapshotExpandableFloatingActionButtonState()
+    rule.setContent {
+      ExpandableFloatingActionButton(
+          state = state,
+          expandedContent = { Text("Expanded") },
+          collapsedContent = { Text("Collapsed") },
+      )
+    }
+    rule.onNodeWithText("Collapsed").performClick()
+    rule.onNodeWithText("Expanded").assertIsDisplayed()
+    assertThat(state.expanded).isTrue()
+  }
+}
--- a/mobile/src/main/java/ch/epfl/sdp/mobile/ui/play/ExpandableFloatingActionButton.kt
+++ b/mobile/src/main/java/ch/epfl/sdp/mobile/ui/play/ExpandableFloatingActionButton.kt
@@ -1,0 +1,239 @@
+package ch.epfl.sdp.mobile.ui.play
+
+import androidx.compose.animation.AnimatedContent
+import androidx.compose.animation.ExperimentalAnimationApi
+import androidx.compose.animation.animateColorAsState
+import androidx.compose.animation.core.animateFloatAsState
+import androidx.compose.animation.with
+import androidx.compose.foundation.BorderStroke
+import androidx.compose.foundation.background
+import androidx.compose.foundation.clickable
+import androidx.compose.foundation.interaction.MutableInteractionSource
+import androidx.compose.foundation.layout.*
+import androidx.compose.foundation.shape.CircleShape
+import androidx.compose.foundation.shape.CornerSize
+import androidx.compose.foundation.shape.RoundedCornerShape
+import androidx.compose.material.*
+import androidx.compose.runtime.*
+import androidx.compose.runtime.saveable.rememberSaveable
+import androidx.compose.ui.Alignment
+import androidx.compose.ui.Modifier
+import androidx.compose.ui.geometry.Size
+import androidx.compose.ui.graphics.Color
+import androidx.compose.ui.unit.Density
+import androidx.compose.ui.unit.Dp
+import androidx.compose.ui.unit.dp
+import androidx.compose.ui.util.lerp
+import ch.epfl.sdp.mobile.ui.*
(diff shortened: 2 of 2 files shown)


matt989253 (approved):
LGTM, thanks! (The correct icons, for this feature as well as the rest of the app, can be added later)

mobile/src/main/java/ch/epfl/sdp/mobile/ui/play/ExpandableFloatingActionButton.kt
```diff
@@ -0,0 +1,240 @@
...
+  }
+
+  @Composable
+  override fun border(expanded: Boolean): State<Color> {
+    return animateColorAsState(if (expanded) Green100 else Green100.copy(alpha = 0f))
+  }
+
+  @Composable
+  override fun contentColor(expanded: Boolean): State<Color> {
+    return animateColorAsState(if (expanded) Green800 else Green100)
```
BadrTad, line 218: [Question] Does 'animateColorAsState' make the `fun contentColor(expanded: Boolean)` returns a slightly different color tint starting from `Green800 `down to `Green100 ` each time the composable gets updated ? Meaning is the returned State<Color> responsible for the triggered updates ?

mobile/src/main/java/ch/epfl/sdp/mobile/ui/play/ExpandableFloatingActionButton.kt
```diff
@@ -0,0 +1,240 @@
...
+  return SnapshotExpandableFloatingActionButtonState(expanded)
+}
+
+/**
+ * An implementation of an [ExpandableFloatingActionButtonState] that delegates to a backing
+ * [MutableState].
+ *
+ * @param state the [MutableState] to which the expanded state is delegated.
+ */
+private class SnapshotExpandableFloatingActionButtonState(
```
BadrTad, line 192: [Question] Why do we chose to delegate expanded state to a `MutableState` ? Why simply not make it itself `MutableState` then use the `by` operand ?

mobile/src/main/java/ch/epfl/sdp/mobile/ui/play/ExpandableFloatingActionButton.kt
```diff
@@ -0,0 +1,240 @@
...
+    colors: ExpandableFloatingActionButtonColors = DefaultExpandableFloatingActionButtonColors,
+    elevation: ButtonElevation = ButtonDefaults.elevation(),
+    interactionSource: MutableInteractionSource = remember { MutableInteractionSource() },
+    expandedCornerSize: Dp = 8.dp,
+    collapsedContent: @Composable RowScope.() -> Unit,
+) {
+  val color by colors.background(state.expanded)
+  val border by colors.border(state.expanded)
+  val contentColor by colors.contentColor(state.expanded)
+  val progress = animateFloatAsState(if (state.expanded) 1f else 0f)
```
BadrTad, line 98: [Question] Does this mean that the progress will take values from 1f to 0f when `expanded ` is `false` ? If so, why is it a `val` and not a `var` ?

mobile/src/androidTest/java/ch/epfl/sdp/mobile/test/ui/play/ExpandableFloatingActionButtonTest.kt
```diff
@@ -0,0 +1,46 @@
...
+
+  @Test
+  fun clickingButton_expandsIt() {
+    rule.setContent {
+      ExpandableFloatingActionButton(
+          expandedContent = { Text("Expanded") },
+          collapsedContent = { Text("Collapsed") },
+      )
+    }
+    rule.onNodeWithText("Collapsed").performClick()
```
barmettlerl, line 29: Nitpick: Can here the LocalizedString be used ?

mobile/src/main/java/ch/epfl/sdp/mobile/ui/play/ExpandableFloatingActionButton.kt
```diff
@@ -0,0 +1,240 @@
...
+@Composable
+fun ExpandableFloatingActionButton(
+    expandedContent: @Composable ColumnScope.() -> Unit,
+    modifier: Modifier = Modifier,
+    state: ExpandableFloatingActionButtonState = rememberExpandableFloatingActionButtonState(),
+    colors: ExpandableFloatingActionButtonColors = DefaultExpandableFloatingActionButtonColors,
+    elevation: ButtonElevation = ButtonDefaults.elevation(),
+    interactionSource: MutableInteractionSource = remember { MutableInteractionSource() },
+    expandedCornerSize: Dp = 8.dp,
+    collapsedContent: @Composable RowScope.() -> Unit,
```
barmettlerl, line 93: Question: What does `RowScope.() -> Unit` mean. Why do you need the RowScope part ?

barmettlerl (approved):

mobile/src/main/java/ch/epfl/sdp/mobile/ui/play/ExpandableFloatingActionButton.kt
```diff
@@ -0,0 +1,240 @@
...
+      elevation = elevation.elevation(enabled = true, interactionSource).value,
+      shape = RoundedCornerShape(corner),
+      modifier = modifier,
+  ) {
+    AnimatedContent(
+        targetState = state.expanded,
+        contentAlignment = Alignment.Center,
+        transitionSpec = { FadeEnterTransition with FadeExitTransition },
+    ) { expanded ->
+      ProvideTextStyle(MaterialTheme.typography.button) {
```
BadrTad, line 114: [Question] Just to make sure that my understanding is correct.  Does the `ProvideTextStyle` composable make sure that everything under its scope will be using the predefined button typography in `MaterialTheme` ? If so why don't we use the default defined style in `PawniesTheme` ?

BadrTad (approved):
Great work. Just a few question to gain a better understanding of Jetpack Compose. Thanks in advance 😀

mobile/src/main/java/ch/epfl/sdp/mobile/ui/play/ExpandableFloatingActionButton.kt
BadrTad: From my understanding` RowScope.() -> Unit` is a Kotlin way to define function types with receivers more [here](https://kotlinlang.org/docs/lambdas.html#function-literals-with-receiver)
If I understood it correctly, the `RowScope` part means that the `collapedContent` composable is drawn from a `RowScope` context.
I'll let @alexandrepiveteau clarify and I hope I am not giving none sense explanation 😁

alexandrepiveteau: That's exactly right @BadrTad 👍 

`RowScope` and `ColumnScope` typically both provide some additional `Modifier` which you can therefore only use when you're in a `RowScope`. An example is `Modifier.align()` (which would not have a well-defined behaviour in layouts which are not rows or columns), so these scopes are used to provide a "type-safe" DSL for building the contents of some rows or columns.

alexandrepiveteau: So long story short, the `RowScope().` receiver is not needed. However, if you want to align some items at the top / bottom, you can apply `Modifier.align()` in the `collapsedContent` composable body.

mobile/src/main/java/ch/epfl/sdp/mobile/ui/play/ExpandableFloatingActionButton.kt
```diff
@@ -0,0 +1,240 @@
...
+private object DefaultExpandableFloatingActionButtonColors : ExpandableFloatingActionButtonColors {
+
+  @Composable
+  override fun background(expanded: Boolean): State<Color> {
+    return animateColorAsState(if (expanded) Beige050 else Green800)
+  }
+
+  @Composable
+  override fun border(expanded: Boolean): State<Color> {
+    return animateColorAsState(if (expanded) Green100 else Green100.copy(alpha = 0f))
```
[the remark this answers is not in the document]
alexandrepiveteau: Not exactly. While both colors have the same `alpha` channel, `Color.Transparent` is defined as `val Transparent = Color(0x00000000)`, so its non-alpha channels are `Black`. This means that animating from one color to the other will actually transition from "transparent black" to "green", and the color will turn gray-ish as it gets more and more transparent.

mobile/src/main/java/ch/epfl/sdp/mobile/ui/play/ExpandableFloatingActionButton.kt
alexandrepiveteau: An important concept of compose is that `State` values which are read by the composition / the layout phase / the drawing phase and which are updated will trigger an invalidation of the composition (respectively layout / drawing phases) if they're updated.

`animateColorAsState` will create a `State` whose value gets smoothly updated at 60 fps when the target changes. **The `State` updates therefore drive the recompositions, relayout or redraws**, depending on how the read value is used.

This also means that once the animation is done (the color is done animating), the `State` does not get updated with new values, and there are no more recompositions.

mobile/src/main/java/ch/epfl/sdp/mobile/ui/play/ExpandableFloatingActionButton.kt
alexandrepiveteau: `ProvideTextStyle` assigns a default `TextStyle` to the `CompositionLocal` used by default by `Text()` composables. When we apply a `PawniesTheme`, we essentially set some values to all the Material Design typography styles, including `MaterialTheme.typography.button`. Therefore, if we retrieve this style later in the composition, it's populated with our `PawniesTheme` typography.

However, this also means we could use the `ExpandableFloatingActionButton` if we're not in a `PawniesTheme` but in another composition which has some other Material Design typography, and the composable would adapt and texts look good by default.

mobile/src/main/java/ch/epfl/sdp/mobile/ui/play/ExpandableFloatingActionButton.kt
```diff
@@ -0,0 +1,240 @@
...
+import ch.epfl.sdp.mobile.ui.FadeExitTransition
+import ch.epfl.sdp.mobile.ui.PawniesColors.Beige050
+import ch.epfl.sdp.mobile.ui.PawniesColors.Green100
+import ch.epfl.sdp.mobile.ui.PawniesColors.Green800
+
+/** The state of an [ExpandableFloatingActionButton] */
+interface ExpandableFloatingActionButtonState {
+
+  /** True iff the button should be currently expanded. */
+  var expanded: Boolean
```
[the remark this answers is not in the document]
alexandrepiveteau: Technically, the `ExpandedFloatinActionButton` composable only has a single property (aka whether it's expanded) for now. Rather than exposing this, It's better to offer this state as a dedicated type (similarly to what we're doing with our screen view models for instance).

You can think of `ExpandableFloatingActionButtonState` as a type alias for `MutableState`, where the `MutableState.value` property is now called `expanded`. That's why we won't expose the `State` if we're actually already in an interface which represents state.

PR #196, opened 2022-03-30
Fix navigation reloading when the profile gets updated

### LINKED ISSUE #195: Updating the username reloads the navigation
When the user is logged in and their username gets updated, the whole application gets reloaded and the current navigation section is lost.

alexandrepiveteau:
This PR fixes #195. A regression test is added, which checks that updating the profile won't change the current navigation section, and a fix is added.

The issue was caused by improper equality checks on `UserState` in the `Crossfade` composable. If the username (or any profile field) was updated, the `UserState` was different and the `Crossfade` would animate to the next state, essentially creating a new body composable.

By specifying an explicit `key` in an `AnimatedContent`, user field updates won't be considered as if a new user had logged in, and navigation reloading will be fixed.

### DIFF
--- a/mobile/src/main/java/ch/epfl/sdp/mobile/state/Navigation.kt
+++ b/mobile/src/main/java/ch/epfl/sdp/mobile/state/Navigation.kt
@@ -1,10 +1,14 @@
 package ch.epfl.sdp.mobile.state
 
-import androidx.compose.animation.Crossfade
+import androidx.compose.animation.AnimatedContent
+import androidx.compose.animation.core.updateTransition
+import androidx.compose.animation.with
 import androidx.compose.runtime.*
 import ch.epfl.sdp.mobile.application.authentication.AuthenticatedUser
 import ch.epfl.sdp.mobile.application.authentication.AuthenticationUser
 import ch.epfl.sdp.mobile.application.authentication.NotAuthenticatedUser
+import ch.epfl.sdp.mobile.ui.FadeEnterTransition
+import ch.epfl.sdp.mobile.ui.FadeExitTransition
 import kotlinx.coroutines.flow.map
 
 /**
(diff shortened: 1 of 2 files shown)


matt989253 (approved):

mobile/src/main/java/ch/epfl/sdp/mobile/state/Navigation.kt
```diff
@@ -27,18 +35,29 @@ fun Navigation() {
 /** An interface representing the current state of the user. */
 private sealed interface UserState {
 
+  /**
+   * A unique key for this [UserState], which indicates that a full-screen transition should occur.
+   */
+  val key: Any
```
BadrTad, line 41: [Question] Why is key of type `Any` ? Wouldn't be more concise if it was an `enum` instead, or simply a `String` ?

BadrTad (approved):

mobile/src/main/java/ch/epfl/sdp/mobile/state/Navigation.kt
alexandrepiveteau: The function `contentKey` from `AnimatedContent` returns an `Any`, so I thought I could simply match its return type. I don't think making it a different type would make the code more concise, but I could indeed enforce that it's a `String`.

However, I would not want future consumers of `UserState` to assume that they'll always get a `key` that's a `String`, so `Any` sounds like a good choice too. What are your thoughts on this ?

BadrTad: What do you mean by future consumers ?

alexandrepiveteau: Some code we write that uses `UserState`

alexandrepiveteau: To me, the fact that a `key` of a `UserState` is a `String` is an implementation detail, so I don't think we should expose a more precise type

BadrTad: Yes sounds logical

PR #197, opened 2022-03-30
Add support for navigating to matches using their id

alexandrepiveteau:
This PR adds match identifiers when displaying a game.


KurohanaJuri (approved):

mobile/src/main/java/ch/epfl/sdp/mobile/state/StatefulGameScreen.kt
```diff
@@ -12,27 +12,24 @@ import ch.epfl.sdp.mobile.ui.game.GameScreen
...
  * @param modifier the [Modifier] for the composable
  */
 @Composable
 fun StatefulGameScreen(
     user: AuthenticatedUser,
+    id: String,
     modifier: Modifier = Modifier,
 ) {
-  val chessFacade = LocalChessFacade.current
+  val facade = LocalChessFacade.current
```
matt989253, line 30: Any reason for this renaming? I think it makes it a bit less clear.

alexandrepiveteau: There's only a single facade used in this composable, so I thought would be easier to read since `facade` is slightly less verbose.

matt989253 (approved):

PR #205, opened 2022-04-01
Profile Game History

Fouad-sys:


mobile/src/main/java/ch/epfl/sdp/mobile/application/authentication/AuthenticatedUser.kt
```diff
@@ -91,4 +93,6 @@ class AuthenticatedUser(
           .whereArrayContains("followers", user.uid)
           .asFlow<ProfileDocument>()
           .map { it.mapNotNull { doc -> doc?.toProfile(this) } }
+
+  val pastGames: Flow<List<ChessMatch>> = emptyFlow()
```
alexandrepiveteau, line 97: The `ChessFacade` is responsible to get the chess games for any specific, so you should use its methods instead.

mobile/src/main/java/ch/epfl/sdp/mobile/application/chess/ChessFacade.kt
```diff
@@ -82,6 +86,48 @@ class ChessFacade(private val auth: Auth, private val store: Store) {
...
+   */
+  suspend fun determineMatchOutcome(match: Match): MatchResult? {
+    val game = match.game.filterNotNull().first()
+    when (game.nextStep) {
+      is NextStep.Checkmate -> MatchResult.Reason.CHECKMATE
+      is NextStep.Stalemate -> Tie
+      else -> Tie
+    }
+    return Tie
+  }
```
alexandrepiveteau, line 103: Please check with @barmettlerl how this can be done nicely. Also, `MatchResult` is a view concept, and should therefore not be used in the `ChessFacade`.

mobile/src/main/java/ch/epfl/sdp/mobile/application/chess/ChessFacade.kt
```diff
@@ -82,6 +86,48 @@ class ChessFacade(private val auth: Auth, private val store: Store) {
...
+
+  /**
+   * Returns a [Flow] of a list of [ChessMatch] of the matches played by the given profile.
+   *
+   * @param profile the profile we want to know their played [ChessMatch]es
+   */
+  fun chessMatches(profile: Profile): Flow<List<ChessMatch>> {
+    val matches = matches(profile)
+    return matches.map { list -> list.mapNotNull { match -> match.toChessMatch(profile.uid) } }
+  }
```
alexandrepiveteau, line 113: This can't be in the `ChessFacade`, since `ChessMatch` is a view concept.

mobile/src/main/java/ch/epfl/sdp/mobile/application/chess/ChessFacade.kt
```diff
@@ -82,6 +86,48 @@ class ChessFacade(private val auth: Auth, private val store: Store) {
...
+      currentUid: String,
+  ): ChessMatch? {
+    val blackID = this.black.filterNotNull().first().uid
+    val whiteID = this.white.filterNotNull().first().uid
+    val adversary = if (blackID == currentUid) whiteID else blackID
+    val result = determineMatchOutcome(this)
+    val game = this.game.first()
+    val moveNum = game.serialize().size
+    return result?.let { ChessMatch(adversary, result, moveNum) }
+  }
```
alexandrepiveteau, line 130: This should not be in the facade, and you should check with @barmettlerl how to make this conversion nicely and in an observable fashion.

mobile/src/main/java/ch/epfl/sdp/mobile/state/StatefulSettingsScreen.kt
```diff
@@ -9,12 +10,14 @@ import ch.epfl.sdp.mobile.ui.setting.SettingScreenState
...
+class AuthenticatedUserProfileScreenState(
+    private val user: AuthenticatedUser,
+    matches: List<ChessMatch>
+) : SettingScreenState {
   override val email = user.email
-  override val pastGamesCount = 0
+  override val pastGamesCount = matches.size
   override val puzzlesCount = 0
-  override val matches = emptyList<ChessMatch>()
+  override val matches = matches
```
alexandrepiveteau, line 20: The `pastGameCount` and `matches` values won't be updated, since they're passed once to the `AuthenticatedUserProfileScreenState` which is `remember`ed.

There are multiple approaches to implementing a ViewModel, and I suggest following a pattern similar to what you can find here : https://github.com/epfl-SDP/android/blob/main/mobile/src/main/java/ch/epfl/sdp/mobile/state/StatefulGameScreen.kt#L99-L127.

Notice how the `Flow` are collected in the ViewModel, using a `CoroutineScope` that's tied to the lifecycle of the composition. When a new element is collected, the associated `MutableState` is updated.

mobile/src/main/java/ch/epfl/sdp/mobile/application/chess/ChessFacade.kt
barmettlerl: Yes, you can merge my branch into yours

mobile/src/main/java/ch/epfl/sdp/mobile/state/StatefulSettingsScreen.kt
```diff
@@ -34,7 +56,61 @@ class AuthenticatedUserProfileScreenState(private val user: AuthenticatedUser) :
...
+}
+
+/**
+ * Returns a [Flow] of a list of [ChessMatch] of the matches played by the given profile.
+ *
+ * @param profile the profile we want to know their played [ChessMatch]es
+ */
+fun chessMatches(profile: Profile, chessFacade: ChessFacade): Flow<List<ChessMatch>> {
+  val matches = chessFacade.matches(profile)
+  return matches.map { list -> list.mapNotNull { match -> match.toChessMatch(profile.uid) } }
```
BadrTad, line 95: [Question ] May the list of `matches` has null elements and thus requires using `mapNotNull` ?

mobile/src/main/java/ch/epfl/sdp/mobile/ui/i18n/English.kt
```diff
@@ -45,6 +45,7 @@ object English : LocalizedStrings {
   override val profileLostByCheckmate = { moves: Int -> "Lost by checkmate after $moves moves" }
   override val profileLostByForfeit = { moves: Int -> "Lost by forfeit after $moves moves" }
   override val profileTieInfo = { moves: Int -> "Tie after $moves moves" }
+  override val profileAgainst = "Against "
```
BadrTad, line 48: There is a extra space next to `"Against "`

mobile/src/main/java/ch/epfl/sdp/mobile/ui/profile/UserScreen.kt
```diff
@@ -29,15 +40,88 @@ fun UserScreen(
...
+@Composable
+fun Match(
+    title: String,
+    subtitle: String,
+    icon: Painter,
+    modifier: Modifier = Modifier,
+) {
+  val titles = title.split(' ', limit = 2)
+  val profileAgainst = titles[0]
+  val profileOpponent = titles[1]
```
BadrTad, line 81: [Question] I don't really understand why is the  `profileAgainst` and `profileOpponent` deduced form the title. Don't they deserve their own arguments ? Plus what it's the difference between them the names are not clear to me.

BadrTad (commented):
Great work ! Just some few remarks.

mobile/src/main/java/ch/epfl/sdp/mobile/ui/i18n/English.kt
Fouad-sys: Indeed, it was meant to separate the word from the opponent's name

mobile/src/main/java/ch/epfl/sdp/mobile/ui/profile/UserScreen.kt
Fouad-sys: Hey Badr,
Yes indeed I was going to have them have separate arguments in the LocalizedStrings (which is why I also added the profileAgainst string value) then decided it may be best to keep the original title argument and split it in half in order to give each part a different color just like in the Figma mockup, while still utilizing the profileMatchTitle function in LocalizedStrings.

mobile/src/main/java/ch/epfl/sdp/mobile/state/StatefulSettingsScreen.kt
Fouad-sys: Probably not, you're right. MapNotNull was just used for safety.

mobile/src/main/java/ch/epfl/sdp/mobile/application/chess/ChessFacade.kt
```diff
@@ -10,6 +10,8 @@ import ch.epfl.sdp.mobile.application.chess.notation.serialize
 import ch.epfl.sdp.mobile.application.toProfile
 import ch.epfl.sdp.mobile.infrastructure.persistence.auth.Auth
 import ch.epfl.sdp.mobile.infrastructure.persistence.store.*
+import ch.epfl.sdp.mobile.ui.social.*
+import java.util.*
```
alexandrepiveteau, line 14: The imports should not be changed unless some new code actually uses them.

mobile/src/main/java/ch/epfl/sdp/mobile/ui/profile/ProfileScreenState.kt
```diff
@@ -15,7 +15,7 @@ interface ProfileScreenState : Person {
   val pastGamesCount: Int
 
   /** List of chess matches */
-  val matches: List<ChessMatch>
+  var matches: List<ChessMatch>
```
alexandrepiveteau, line 18: The consumers of `ProfileScreenState` should not be able to set the value of the `matches`, so this should remain a `val`.

```suggestion
  val matches: List<ChessMatch>
```

mobile/src/main/java/ch/epfl/sdp/mobile/ui/profile/ProfileScreen.kt
```diff
@@ -19,6 +19,7 @@ import androidx.compose.ui.unit.Dp
 import androidx.compose.ui.unit.dp
 import ch.epfl.sdp.mobile.state.LocalLocalizedStrings
 import ch.epfl.sdp.mobile.state.toColor
+import ch.epfl.sdp.mobile.ui.social.*
```
alexandrepiveteau, line 22: The imports should not be changed unless some new code actually uses them.

mobile/src/androidTest/java/ch/epfl/sdp/mobile/test/state/StatefulProfileScreenTest.kt
```diff
@@ -0,0 +1,48 @@
...
+import ch.epfl.sdp.mobile.test.infrastructure.persistence.store.document
+import kotlinx.coroutines.test.runTest
+import org.junit.Rule
+import org.junit.Test
+
+class StatefulProfileScreenTest {
+  @get:Rule val rule = createComposeRule()
+
+  @Test
+  fun Given_statefulProfileScreen_When_profileHasPastGames_Then_theyAreDisplayedOnScreen() {
```
alexandrepiveteau, line 23:
```suggestion
  fun given_statefulProfileScreen_when_profileHasPastGames_then_theyAreDisplayedOnScreen() {
```

mobile/src/androidTest/java/ch/epfl/sdp/mobile/test/state/StatefulSettingsScreenTest.kt
```diff
@@ -0,0 +1,54 @@
...
+import kotlinx.coroutines.flow.first
+import kotlinx.coroutines.test.runTest
+import org.junit.Rule
+import org.junit.Test
+
+class StatefulSettingsScreenTest {
+  @get:Rule val rule = createComposeRule()
+
+  @Test
+  fun Given_statefulSettingsScreen_When_profileHasPastGames_Then_theyAreDisplayedOnScreen() {
```
alexandrepiveteau, line 26:
```suggestion
  fun given_statefulSettingsScreen_when_profileHasPastGames_then_theyAreDisplayedOnScreen() {
```

mobile/src/androidTest/java/ch/epfl/sdp/mobile/test/ui/profile/ProfileScreenTest.kt
```diff
@@ -22,7 +22,7 @@ class ProfileScreenTest {
 
   @get:Rule val rule = createComposeRule()
 
-  open class TestProfileScreenState(override val matches: List<ChessMatch>) : ProfileScreenState {
+  open class TestProfileScreenState(override var matches: List<ChessMatch>) : ProfileScreenState {
```
alexandrepiveteau, line 25: Should be a `val`.

mobile/src/androidTest/java/ch/epfl/sdp/mobile/test/ui/setting/SettingScreenTest.kt
```diff
@@ -21,9 +21,9 @@ class SettingScreenTest {
 
   @get:Rule val rule = createComposeRule()
 
-  open class TestSettingScreenState(override val matches: List<ChessMatch>) : SettingScreenState {
+  open class TestSettingScreenState(override var matches: List<ChessMatch>) : SettingScreenState {
```
alexandrepiveteau, line 24: Should be a `val`.

mobile/src/androidTest/java/ch/epfl/sdp/mobile/test/ui/setting/SettingScreenTest.kt
```diff
@@ -21,9 +21,9 @@ class SettingScreenTest {
...
     override val email = "example@epfl.ch"
-    override val pastGamesCount = 10
+    override var pastGamesCount = 10
```
alexandrepiveteau, line 26: Should be a `val`.

mobile/src/main/java/ch/epfl/sdp/mobile/state/StatefulProfileScreen.kt
```diff
@@ -36,8 +34,14 @@ fun StatefulProfileScreen(
...
   val socialFacade = LocalSocialFacade.current
+  val chessFacade = LocalChessFacade.current
   val profile by remember(socialFacade, uid) { socialFacade.profile(uid) }.collectAsState(null)
-  val state = profile?.let { FetchedUserProfileScreenState(it) }
+  val matches =
+      remember(chessFacade, profile) { profile?.let { chessMatches(it, chessFacade) } }
+          ?.collectAsState(emptyList())
+          ?.value
+          ?: emptyList()
+  val state = profile?.let { FetchedUserProfileScreenState(it, matches) }
```
alexandrepiveteau, line 44: I think you should use the same pattern as here (https://github.com/epfl-SDP/android/blob/main/mobile/src/main/java/ch/epfl/sdp/mobile/state/StatefulFollowingScreen.kt#L91-L151) to manage the state using the facades and an eventual `CoroutineScope`.

mobile/src/main/java/ch/epfl/sdp/mobile/state/StatefulProfileScreen.kt
```diff
@@ -36,8 +34,14 @@ fun StatefulProfileScreen(
...
   if (state != null) {
```
alexandrepiveteau, line 45: This is some legacy code, but having a `null` `state` means something went terribly wrong, and we should get rid of it.

mobile/src/main/java/ch/epfl/sdp/mobile/ui/setting/SettingScreenState.kt
```diff
@@ -11,13 +11,13 @@ interface SettingScreenState : Person {
   val email: String
 
   /** Number of past games */
-  val pastGamesCount: Int
+  var pastGamesCount: Int
```
alexandrepiveteau, line 14:
```suggestion
  val pastGamesCount: Int
```

mobile/src/main/java/ch/epfl/sdp/mobile/ui/setting/SettingScreenState.kt
```diff
@@ -11,13 +11,13 @@ interface SettingScreenState : Person {
...
 
   /** Number of puzzles */
   val puzzlesCount: Int
 
   /** List of chess matches */
-  val matches: List<ChessMatch>
+  var matches: List<ChessMatch>
```
alexandrepiveteau, line 20:
```suggestion
  val matches: List<ChessMatch>
```

mobile/src/main/java/ch/epfl/sdp/mobile/ui/setting/SettingScreen.kt
```diff
@@ -33,30 +34,37 @@ import ch.epfl.sdp.mobile.ui.profile.rememberSettingTabBarState
...
+                  modifier = Modifier.fillMaxWidth(),
+                  elevation = elevation,
+              )
+            },
+            matches = state.matches,
+            lazyColumnState = lazyColumnState,
+            modifier = modifier.fillMaxSize(),
+            contentPadding = totalPadding,
+            onMatchClick = {})
+      })
```
alexandrepiveteau, line 67: What is this `Scaffold` used for ?

mobile/src/main/java/ch/epfl/sdp/mobile/ui/profile/UserScreen.kt
alexandrepiveteau: I 100% agree with @BadrTad. If the format of the title changes, this will go terribly wrong.

alexandrepiveteau (changes requested):
I'm not sure I understand all of the choices that were made here. To me, it looks like there's a lot of duplication with #212. Also, the logic should likely be identical between the settings screen and the profile screen.

I also believe some features are missing - it's not possible to click on a match to display it.

mobile/src/main/java/ch/epfl/sdp/mobile/ui/profile/UserScreen.kt
Fouad-sys: Fixed in 87b7aea

mobile/src/main/java/ch/epfl/sdp/mobile/application/chess/ChessFacade.kt
Fouad-sys: Fixed in 87b7aea

mobile/src/main/java/ch/epfl/sdp/mobile/ui/setting/SettingScreen.kt
Fouad-sys: Removed and replaced with UserScreen in e2d401e1c120ebfdd323d3d2971f29cb337d812b

Fouad-sys: Removed and replaced with UserScreen in e2d401e1c120ebfdd323d3d2971f29cb337d812b

mobile/src/main/java/ch/epfl/sdp/mobile/ui/profile/Match.kt
```diff
@@ -1,47 +1,66 @@
...
 import ch.epfl.sdp.mobile.ui.social.*
 
 /**
- * Composes a Match log using a match [title], [subtitle] and an [icon]
- * @param match a chess Match
- * @param onClick callback function if clicked on item
+ * Composes a Match log using a match [adversary], [subtitle] and an [icon].
+ *
+ * @param match the [ChessMatch] to compose.
+ * @param icon match icon.
```
KurohanaJuri, line 23: Miss `icon` param

mobile/src/main/java/ch/epfl/sdp/mobile/state/StatefulSettingsScreen.kt
```diff
@@ -1,21 +1,37 @@
...
 import ch.epfl.sdp.mobile.application.authentication.AuthenticatedUser
+import ch.epfl.sdp.mobile.application.chess.ChessFacade
 import ch.epfl.sdp.mobile.ui.setting.SettingScreenState
 import ch.epfl.sdp.mobile.ui.setting.SettingsScreen
 import ch.epfl.sdp.mobile.ui.social.ChessMatch
 import ch.epfl.sdp.mobile.ui.social.Person
+import kotlinx.coroutines.CoroutineScope
+import kotlinx.coroutines.launch
 
 class AuthenticatedUserProfileScreenState(
```
barmettlerl, line 15: Might Require some Documentation

mobile/src/main/java/ch/epfl/sdp/mobile/ui/profile/UserScreen.kt
```diff
@@ -29,15 +33,18 @@ fun UserScreen(
...
       state = lazyColumnState,
       verticalArrangement = Arrangement.Top,
       horizontalAlignment = Alignment.CenterHorizontally,
       modifier = modifier,
+      contentPadding = contentPadding,
   ) {
     item { header() }
     stickyHeader { profileTabBar() }
-    items(matches) { match -> Match(match, { onMatchClick(match) }) }
+    items(matches) { match -> Match(match, ChessIcons.WhiteKing, onMatchClick) }
```
barmettlerl, line 48: Should the color of the white king change whenever I'm the black player or does it stays always white ?

barmettlerl (approved):

mobile/src/main/java/ch/epfl/sdp/mobile/ui/profile/UserScreen.kt
Fouad-sys: No I think it is a default icon on figma

mobile/src/main/java/ch/epfl/sdp/mobile/state/StatefulSettingsScreen.kt
Fouad-sys: Fixed in 9337db8ae78de9150eed15e056e78fb9165dd16a

mobile/src/main/java/ch/epfl/sdp/mobile/state/StatefulProfileScreen.kt
Fouad-sys: Fixed in 9337db8ae78de9150eed15e056e78fb9165dd16a

mobile/src/main/java/ch/epfl/sdp/mobile/ui/profile/Match.kt
KurohanaJuri: My bad, I didn't see the icon parameter the first time I read it

KurohanaJuri (approved):

mobile/src/main/java/ch/epfl/sdp/mobile/state/StatefulProfileScreen.kt
```diff
@@ -27,16 +49,28 @@ class FetchedUserProfileScreenState(
...
+        override val name: String = ""
+        override val backgroundColor: Profile.Color = Profile.Color.Default
+        override val uid: String = ""
+        override val followed: Boolean = false
+      }
+  val fetchedProfile by
+      remember(socialFacade, uid) { socialFacade.profile(uid) }.collectAsState(null)
+  val profile = fetchedProfile ?: emptyProfile
+  val scope = rememberCoroutineScope()
+  val state = remember(profile) { FetchedUserProfileScreenState(profile, chessFacade, scope) }
```
alexandrepiveteau, line 74: Bug : you must specify some dependencies. If the `ChessFacade` or the `CoroutineScope` were to be changed, you'd need to re-create the `FetchedUserProfileScreenState`.

```suggestion
  val state = remember(profile, chessFacade, scope) { FetchedUserProfileScreenState(profile, chessFacade, scope) }
```

alexandrepiveteau (changes requested):
Great work, however there are still some details to address.

Aside from the bugs in the usage of `remember()`, you should introduce a more precise inheritance hierarchy in the states to avoid code duplication.

BadrTad (commented):
Great work ! Just a some few remarks

mobile/src/androidTest/java/ch/epfl/sdp/mobile/test/state/StatefulSettingsScreenTest.kt
```diff
@@ -0,0 +1,56 @@
...
+      authFacade.signInWithEmail("email@example.org", "password")
+      val user = authFacade.currentUser.filterIsInstance<AuthenticatedUser>().first()
+
+      val strings =
+          rule.setContentWithLocalizedStrings {
+            ProvideFacades(authFacade, socialFacade, chessFacade) {
+              StatefulSettingsScreen(user, {})
+            }
+          }
+      rule.onNodeWithText(strings.profileMatchTitle("B")).assertExists()
```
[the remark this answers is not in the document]
Fouad-sys: Yes, but since there is only one game here, it is best to check the description of the game

mobile/src/androidTest/java/ch/epfl/sdp/mobile/test/state/StatefulProfileScreenTest.kt
```diff
@@ -0,0 +1,68 @@
...
+      rule.onNodeWithText(strings.profileMatchTitle("B")).assertExists()
+    }
+  }
+
+  @Test
+  fun given_statefulProfileScreen_when_profileStateIsNull_then_aNullProfileIsDisplayed() {
+    runTest {
+      val auth = buildAuth { user("email@example.org", "password", "1") }
+      val store = buildStore {
+        collection("users") { document("1", ProfileDocument("1", name = "B")) }
```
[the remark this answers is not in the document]
Fouad-sys: Yes, but it is best to have a connected user for the test

mobile/src/main/java/ch/epfl/sdp/mobile/state/StatefulProfileScreen.kt
```diff
@@ -1,42 +1,39 @@
...
+class StatefulProfileScreen(
+    user: Profile,
+    private val chessFacade: ChessFacade,
+    private val scope: CoroutineScope,
+) : ProfileScreenState, Person by ProfileAdapter(user) {
+  override val email: String = ""
+  override var matches by mutableStateOf(emptyList<ChessMatch>())
+    private set
+  override var pastGamesCount = matches.size
+    private set
```
alexandrepiveteau, line 31: Bug : if the matches list gets updated, the count won't be updated.

```suggestion
  override val pastGamesCount
    get() = matches.size
```

mobile/src/main/java/ch/epfl/sdp/mobile/state/StatefulSettingsScreen.kt
```diff
@@ -1,27 +1,33 @@
...
+ * @param onEditProfileNameClickAction Callable lambda to navigate to the profile Edit popup
  */
-class SettingsScreenStateImpl(
-    user: AuthenticatedUser,
-    onEditProfileNameClickAction: State<() -> Unit>
-) : SettingScreenState, Person by ProfileAdapter(user) {
+class AuthenticatedUserProfileScreenState(
+    private val user: AuthenticatedUser,
+    private val chessFacade: ChessFacade,
+    private val scope: CoroutineScope,
```
alexandrepiveteau, line 26: Could these fields not be `val` ?

mobile/src/main/java/ch/epfl/sdp/mobile/state/StatefulProfileScreen.kt
```diff
@@ -1,42 +1,39 @@
...
-  val profile by remember(socialFacade, uid) { socialFacade.profile(uid) }.collectAsState(null)
-  val state = profile?.let { FetchedUserProfileScreenState(it) }
-  if (state != null) {
-    ProfileScreen(state, modifier)
+class StatefulProfileScreen(
+    user: Profile,
+    private val chessFacade: ChessFacade,
+    private val scope: CoroutineScope,
+) : ProfileScreenState, Person by ProfileAdapter(user) {
+  override val email: String = ""
```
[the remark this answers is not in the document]
Fouad-sys: I had that intuition as well

mobile/src/main/java/ch/epfl/sdp/mobile/ui/profile/ProfileScreen.kt
```diff
@@ -66,7 +71,7 @@ fun ProfileHeader(state: ProfileScreenState, modifier: Modifier = Modifier) {
     ProfilePicture(state)
     Column(horizontalAlignment = Alignment.CenterHorizontally) {
       Text(state.name, style = MaterialTheme.typography.h5)
-      Text(state.email, style = MaterialTheme.typography.subtitle2)
+      Text("", style = MaterialTheme.typography.subtitle2)
```
alexandrepiveteau, line 74: This seems a bit off to me. Why is this `Text` composable always empty ?

Fouad-sys: Fixed in e2d18a79d8d6c16fbd877e58ae775b417ef6e1cb

mobile/src/androidTest/java/ch/epfl/sdp/mobile/test/state/StatefulProfileScreenTest.kt
```diff
@@ -0,0 +1,70 @@
...
+      }
+      val authFacade = AuthenticationFacade(auth, store)
+      val socialFacade = SocialFacade(auth, store)
+      val chessFacade = ChessFacade(auth, store)
+      rule.setContentWithLocalizedStrings {
+        ProvideFacades(authFacade, socialFacade, chessFacade) { StatefulVisitedProfileScreen("2") }
+      }
+
+      rule.onAllNodesWithText("").assertCountEquals(2)
```
alexandrepiveteau, line 67: It's a bad practice to rely on the number of nodes for assertions. Moreover, it looks like a bad practice to rely on the nodes being empty.

What are you trying to test exactly ? Could some logic be extracted to the `FetchedUserProfileScreenState` and tested from there ?

alexandrepiveteau (changes requested):

alexandrepiveteau (approved):
Great work 🎉

matt989253 (approved):
Very nice! 🍾

PR #212, opened 2022-04-03
Display active games

barmettlerl:


mobile/src/main/java/ch/epfl/sdp/mobile/ui/play/PlayScreen.kt
```diff
@@ -24,8 +31,30 @@ import ch.epfl.sdp.mobile.ui.PawniesIcons
 fun PlayScreen(
     state: PlayScreenState,
     modifier: Modifier = Modifier,
-    contentPadding: PaddingValues = PaddingValues(),
+    contentPadding: PaddingValues = PaddingValues()
 ) {
+  val lazyColumnState = rememberLazyListState()
```
alexandrepiveteau, line 36: You don't seem to be using this hoisted state except for passing it to the `LazyColumn`, so I suggest you don't explicitly manage it.

mobile/src/main/java/ch/epfl/sdp/mobile/ui/play/PlayScreen.kt
```diff
@@ -24,8 +31,30 @@ import ch.epfl.sdp.mobile.ui.PawniesIcons
...
+    Text(
+        text = strings.playOnlineGames,
+        color = MaterialTheme.colors.primary,
+        style = MaterialTheme.typography.h4,
+        modifier = Modifier.padding(horizontal = 16.dp, vertical = 32.dp))
+    LazyColumn(
+        state = lazyColumnState,
+        verticalArrangement = Arrangement.Top,
+        horizontalAlignment = Alignment.CenterHorizontally,
+        modifier = modifier,
```
alexandrepiveteau, line 48: Bug : You must not use `modifier` if this is not the top-level composable.
```suggestion
        modifier = Modifier,
```

mobile/src/main/java/ch/epfl/sdp/mobile/ui/play/PlayScreenState.kt
```diff
@@ -1,10 +1,12 @@
...
+import ch.epfl.sdp.mobile.ui.social.ChessMatch
+
 /** Interface that represents state of the PlayScreen */
 interface PlayScreenState {
 
   /* Callable upon actioning button */
   val onNewGameClick: () -> Unit
 
-  /* TODO: add necessary attributes: games log... */
+  val matches: List<ChessMatch>
```
alexandrepiveteau, line 11: Missing documentation.

mobile/src/main/java/ch/epfl/sdp/mobile/ui/profile/Match.kt
```diff
@@ -0,0 +1,60 @@
...
+    icon: ImageVector,
+    modifier: Modifier = Modifier,
+) {
+  ListItem(
+      modifier = modifier,
+      icon = { Icon(icon, null) },
+      text = { Text(title) },
+      secondaryText = { Text(subtitle) },
+  )
+}
```
alexandrepiveteau, line 37: Good idea to extract this in its own file. However, I don't see the point of this composable, given that the caller has to pass in the subtitle too. What would you think of refactoring it (alongside with `chooseSubtitle` to have a signature a bit more like this one ?

```kotlin
@Composable
fun Match(
  match: ChessMatch,
) {
  // ...
}
```

This way, the function `chooseSubtitle` could be kept `private` and the composable have an actual utility.

mobile/src/main/java/ch/epfl/sdp/mobile/state/StatefulPlayScreen.kt
```diff
@@ -1,34 +1,141 @@
...
  */
-/* FIXME: add more useful args : user... */
-class Player(private val newGame: () -> Unit) : PlayScreenState {
-  /* TODO: override other attributes */
-  override val onNewGameClick = newGame
+inline fun <A, B> Loadable<A>.map(f: (A) -> B): Loadable<B> =
+    when (this) {
+      is Loadable.Loaded -> loaded(f(value))
+      Loadable.Loading -> loading()
+    }
```
alexandrepiveteau, line 38: I don't think these should be exposed as `public`. These classes are just an implementation detail of the `PlayScreenStateImpl`.

mobile/src/main/java/ch/epfl/sdp/mobile/state/StatefulPlayScreen.kt
```diff
@@ -1,34 +1,141 @@
...
+      fetchForUser(user, facade).collect { list ->
+        matches =
+            list.map {
+              ChessMatch(
+                  adversary = if (user.uid == it.blackId) it.blackName else it.whiteName,
+                  matchResult = Tie,
+                  numberOfMoves = it.movesCount,
+              )
+            }
+      }
```
alexandrepiveteau, line 109: Maybe this mapping could be extracted to a dedicated function ?

mobile/src/main/java/ch/epfl/sdp/mobile/state/StatefulPlayScreen.kt
```diff
@@ -1,34 +1,141 @@
...
+    )
+  }
+}
+
+private class PlayScreenStateImpl(
+    onNewGameClick: State<() -> Unit>,
+    user: AuthenticatedUser,
+    facade: ChessFacade,
+    scope: CoroutineScope,
+) : PlayScreenState {
```
alexandrepiveteau, line 93: Missing some documentation.

mobile/src/main/java/ch/epfl/sdp/mobile/state/StatefulPlayScreen.kt
```diff
@@ -1,34 +1,141 @@
...
+ * Extract the value out of the Loadable type
+ * @param lazyBlock function which returns alternative value
+ */
+inline fun <A> Loadable<A>.orElse(lazyBlock: () -> A): A =
+    when (this) {
+      is Loadable.Loaded -> value
+      Loadable.Loading -> lazyBlock()
+    }
+
+private data class MatchInfo(
```
alexandrepiveteau, line 50: Missing documentation.

mobile/src/main/java/ch/epfl/sdp/mobile/state/StatefulPlayScreen.kt
```diff
@@ -1,34 +1,141 @@
...
+    val whiteId: String,
+    val blackId: String,
+    val whiteName: String,
+    val blackName: String,
+)
+
+private fun fetchForUser(
+    currentUser: AuthenticatedUser,
+    facade: ChessFacade,
+): Flow<List<MatchInfo>> =
```
alexandrepiveteau, line 62: Missing documentation.

mobile/src/main/java/ch/epfl/sdp/mobile/state/StatefulPlayScreen.kt
```diff
@@ -1,34 +1,141 @@
...
+    facade.matches(currentUser).map { m -> m.map { info(it) } }.flatMapLatest { matches ->
+      combine(matches) { it.toList() }
+    }
+
+private fun info(match: Match): Flow<MatchInfo> {
```
alexandrepiveteau, line 67: Missing documentation.

mobile/src/main/java/ch/epfl/sdp/mobile/state/StatefulHome.kt
```diff
@@ -82,7 +82,7 @@ fun StatefulHome(
       }
       composable(PlayRoute) {
         StatefulPlayScreen(
-            { controller.navigate(PrepareGameRoute) }, Modifier.fillMaxSize(), paddingValues)
+            user, { controller.navigate(PrepareGameRoute) }, Modifier.fillMaxSize(), paddingValues)
```
alexandrepiveteau, line 85: I know it wasn't previously named, but I think naming arguments would make it much easier to understand what the lambda is for.

alexandrepiveteau (commented):
Nice work, thanks !

Aside from the code coverage which may need to be increased a bit more, I think you still have to implement navigation to games by making matches clickable. Will it be part of this PR or some ulterior work ?

mobile/src/main/java/ch/epfl/sdp/mobile/state/StatefulPlayScreen.kt
barmettlerl: It's a private class

mobile/src/main/java/ch/epfl/sdp/mobile/state/StatefulPlayScreen.kt
barmettlerl: Need to be public in order to write tests for it

barmettlerl:
by accident closed it -> will reopen it

mobile/src/main/java/ch/epfl/sdp/mobile/state/StatefulPlayScreen.kt
alexandrepiveteau: Because it's an implementation detail, maybe the tests should not target the functions exposed in Loadable directly, but it should be tested indirectly when you check the behavior of the viewmodel.

mobile/src/main/java/ch/epfl/sdp/mobile/state/StatefulPlayScreen.kt
```diff
@@ -1,34 +1,164 @@
...
+
+/**
+ * Helper function for createChessMatch
+ * @param match [MatchInfo] intermediate datatype
+ * @param user authenticated user
+ */
+private fun createChessMatch(match: MatchInfo, user: AuthenticatedUser): ChessMatch =
+    ChessMatch(
+        adversary = if (user.uid == match.blackId) match.blackName else match.whiteName,
+        matchResult = Tie,
```
matt989253, line 133: A chess match probably isn't always tied

matt989253 (changes requested):
If a match with the user's id but no id for his opponent is present in the `Store`, the app crashes. This probably shouldn't happen.

mobile/src/main/java/ch/epfl/sdp/mobile/ui/social/ChessMatch.kt
```diff
@@ -1,6 +1,11 @@
 package ch.epfl.sdp.mobile.ui.social
 
-class ChessMatch(val adversary: String, val matchResult: MatchResult, val numberOfMoves: Int)
+class ChessMatch(
+    val uid: String,
```
alexandrepiveteau, line 4: A `ChessMatch` should not have a `uid`, but instead you should use the same pattern as the generic `Person` (https://github.com/epfl-SDP/android/blob/main/mobile/src/main/java/ch/epfl/sdp/mobile/ui/social/SocialScreenState.kt#L16-L56) or the generic `Piece` (https://github.com/epfl-SDP/android/blob/main/mobile/src/main/java/ch/epfl/sdp/mobile/ui/game/ChessBoardState.kt#L20-L130).

mobile/src/main/java/ch/epfl/sdp/mobile/state/StatefulHome.kt
```diff
@@ -62,6 +63,10 @@ fun StatefulHome(
     controller.navigate("$ProfileRoute/${person.uid}")
   }
 
+  val onGameItemClick: (match: ChessMatch) -> Unit = { match ->
+    controller.navigate("$GameRoute/${match.uid}")
+  }
```
alexandrepiveteau, line 68: Shouldn't be a `ChessMatch` but your ViewModel-specific implementation of it.

mobile/src/main/java/ch/epfl/sdp/mobile/state/StatefulPlayScreen.kt
```diff
@@ -1,34 +1,189 @@
...
+/** An Loadable Union Type to differentiate between loaded and state */
+sealed interface Loadable<out T> {
+  object Loading : Loadable<Nothing>
+  data class Loaded<out T>(val value: T) : Loadable<T>
+
+  companion object {
+    fun loading(): Loadable<Nothing> = Loading
+    fun <T> loaded(value: T): Loadable<T> = Loaded(value)
+  }
+}
```
alexandrepiveteau, line 30: I think this should be made `private`.

mobile/src/main/java/ch/epfl/sdp/mobile/state/StatefulPlayScreen.kt
```diff
@@ -1,34 +1,189 @@
...
+      is NextStep.Stalemate -> { _ -> Tie }
+      is NextStep.MovePiece -> { userColor -> if (this.turn == userColor) YourTurn else OtherTurn }
+      is NextStep.Checkmate -> { userColor ->
+            if (userColor == this.winner) {
+              Win(MatchResult.Reason.CHECKMATE)
+            } else {
+              Loss(MatchResult.Reason.CHECKMATE)
+            }
+          }
+    }
```
alexandrepiveteau, line 66: Any reason for this function to be `inline` ? It doesn't seem to have lambda arguments or need reified generics.

Nice use of a function return type though !

mobile/src/main/java/ch/epfl/sdp/mobile/ui/play/PlayScreenState.kt
```diff
@@ -1,10 +1,16 @@
...
+
 /** Interface that represents state of the PlayScreen */
 interface PlayScreenState {
 
   /* Callable upon actioning button */
   val onNewGameClick: () -> Unit
 
-  /* TODO: add necessary attributes: games log... */
+  /** Action to execute when clicked on match item in list */
+  val onGameItemClick: (ChessMatch) -> Unit
```
alexandrepiveteau, line 12: Should be `fun onGameItemClick(match: ChessMatch)` instead.

mobile/src/main/java/ch/epfl/sdp/mobile/ui/profile/Match.kt
```diff
@@ -0,0 +1,70 @@
...
+ * @param title match title
+ * @param subtitle match subtitle info
+ * @param icon match icon
+ * @param modifier the [Modifier] for this composable.
+ */
+@OptIn(ExperimentalMaterialApi::class)
+@Composable
+fun Match(
+    match: ChessMatch,
+    onGameItemClick: (ChessMatch) -> Unit,
```
alexandrepiveteau, line 27: Why not just `onClick: () -> Unit` ? The caller knows which match is associated with the composable.

mobile/src/main/java/ch/epfl/sdp/mobile/ui/profile/Match.kt
```diff
@@ -0,0 +1,70 @@
...
+    modifier: Modifier = Modifier,
+) {
```
alexandrepiveteau, line 29: Mismatch between the documentation and the parameters.

mobile/src/main/java/ch/epfl/sdp/mobile/ui/profile/UserScreen.kt
```diff
@@ -30,11 +25,10 @@ fun UserScreen(
     header: @Composable () -> Unit,
     profileTabBar: @Composable () -> Unit,
     matches: List<ChessMatch>,
+    onGameItemClick: (ChessMatch) -> Unit,
```
alexandrepiveteau, line 28: Maybe `onMatchClick` would be a more consistent naming with our existing `matches` parameter.

mobile/src/main/java/ch/epfl/sdp/mobile/state/StatefulPlayScreen.kt
```diff
@@ -1,34 +1,189 @@
...
+/**
+ * Implementation of the PlayscreenState
+ * @param onNewGameClick callback function for the new game button
+ * @param user authenticated user
+ * @param facade [ChessFacade] to fetch the matches
+ * @param scope for coroutines
+ */
+private class PlayScreenStateImpl(
+    onNewGameClick: State<() -> Unit>,
+    override val onGameItemClick: (ChessMatch) -> Unit,
```
alexandrepiveteau, line 129: Like `onNewGameClick`, you should pass this as a `State` and use a `rememberUpdatedState` when creating the `PlayScreenStateImpl`

mobile/src/main/java/ch/epfl/sdp/mobile/ui/play/PlayScreen.kt
```diff
@@ -24,8 +28,22 @@ import ch.epfl.sdp.mobile.ui.PawniesIcons
...
+    Text(
+        text = strings.playOnlineGames,
+        color = MaterialTheme.colors.primary,
+        style = MaterialTheme.typography.h4,
+        modifier = Modifier.padding(horizontal = 16.dp, vertical = 32.dp))
+    LazyColumn(
+        verticalArrangement = Arrangement.Top,
+        horizontalAlignment = Alignment.CenterHorizontally,
+        modifier = Modifier,
+    ) { items(state.matches) { match -> Match(match, state.onGameItemClick) } }
```
alexandrepiveteau, line 44: Maybe these matches would benefit to be keyed, using a pattern similar to this : https://github.com/epfl-SDP/android/blob/main/mobile/src/main/java/ch/epfl/sdp/mobile/ui/social/SocialScreen.kt#L44.

alexandrepiveteau (commented):
Nice work on the functionality ! I think a few adjustments are still needed to avoid the abstraction leak from `uid` into `ChessMatch`, but otherwise it's awesome seeing some (real) games !

matt989253:
New tests that you added (not modified) should follow the new test naming convention that was decided together.

alexandrepiveteau (changes requested):
The coverage is still a bit low (compared to where it could be), and this lets some bugs slip through (for instance, nothing happening when one clicks a match on a user profile).

Additionally, the app crashes if the user has a local game, since a single game is present twice in the `LazyColumn`. This could be solved by changing the last line of `ChessFacade.matches()` to

```kotlin
return combine(gamesAsWhite, gamesAsBlack) { (a, b) -> a.union(b).sortedBy { it.id } }
```

Is this something you could please include in this PR ?

mobile/src/main/java/ch/epfl/sdp/mobile/ui/play/PlayScreenState.kt
```diff
@@ -1,10 +1,16 @@
...
 /** Interface that represents state of the PlayScreen */
-interface PlayScreenState {
+interface PlayScreenState<M : ChessMatch> {
 
   /* Callable upon actioning button */
   val onNewGameClick: () -> Unit
 
-  /* TODO: add necessary attributes: games log... */
+  /** Action to execute when clicked on match item in list */
+  val onGameItemClick: (M) -> Unit
```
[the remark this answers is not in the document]
barmettlerl: Why should this be a method ?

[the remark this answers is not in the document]
barmettlerl: And why then we don't do that for `onNewGameClick` ?

mobile/src/main/java/ch/epfl/sdp/mobile/ui/play/PlayScreen.kt
```diff
@@ -13,19 +16,37 @@ import ch.epfl.sdp.mobile.state.LocalLocalizedStrings
...
+  val strings = LocalLocalizedStrings.current
+  Column(modifier) {
+    Text(
+        text = strings.playOnlineGames,
+        color = MaterialTheme.colors.primary,
+        style = MaterialTheme.typography.h4,
+        modifier = Modifier.padding(horizontal = 16.dp, vertical = 32.dp))
+    LazyColumn(
+        verticalArrangement = Arrangement.Top,
+        horizontalAlignment = Alignment.CenterHorizontally,
```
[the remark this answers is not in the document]
barmettlerl: Not sure what you mean ?

mobile/src/main/java/ch/epfl/sdp/mobile/ui/play/PlayScreenState.kt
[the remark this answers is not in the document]
alexandrepiveteau: It should have been done for `onNewGameClick` too.

mobile/src/main/java/ch/epfl/sdp/mobile/ui/profile/Match.kt
```diff
@@ -0,0 +1,69 @@
...
+import ch.epfl.sdp.mobile.state.LocalLocalizedStrings
+import ch.epfl.sdp.mobile.ui.PawniesIcons
+import ch.epfl.sdp.mobile.ui.SectionSocial
+import ch.epfl.sdp.mobile.ui.i18n.LocalizedStrings
+import ch.epfl.sdp.mobile.ui.social.*
+
+/**
+ * Composes a Match log using a match [title], [subtitle] and an [icon]
+ * @param match a chess Match
+ * @param onMatchClick callback function if clicked on item
```
alexandrepiveteau, line 19: Small inconsistency following renaming

```suggestion
 * @param onClick callback function if clicked on item
```

mobile/src/main/java/ch/epfl/sdp/mobile/state/StatefulPlayScreen.kt
```diff
@@ -1,34 +1,204 @@
...
+  companion object {
+    fun loading(): Loadable<Nothing> = Loading
+    fun <T> loaded(value: T): Loadable<T> = Loaded(value)
+  }
+}
+/**
+ * A map for the Loadable type
+ * @param f callback function for map
+ */
+private inline fun <A, B> Loadable<A>.map(f: (A) -> B): Loadable<B> =
```
alexandrepiveteau, line 68:
```suggestion
private fun <A, B> Loadable<A>.map(f: (A) -> B): Loadable<B> =
```

mobile/src/main/java/ch/epfl/sdp/mobile/state/StatefulPlayScreen.kt
```diff
@@ -1,34 +1,204 @@
...
+    when (this) {
+      is Loadable.Loaded -> loaded(f(value))
+      Loadable.Loading -> loading()
+    }
+
+/**
+ * Extract the value out of the Loadable type
+ * @param lazyBlock function which returns alternative value
+ */
+private inline fun <A> Loadable<A>.orElse(lazyBlock: () -> A): A =
```
alexandrepiveteau, line 78:
```suggestion
private fun <A> Loadable<A>.orElse(lazyBlock: () -> A): A =
```

mobile/src/main/java/ch/epfl/sdp/mobile/ui/play/PlayScreen.kt
```diff
@@ -13,19 +16,36 @@ import ch.epfl.sdp.mobile.state.LocalLocalizedStrings
...
+import ch.epfl.sdp.mobile.ui.profile.Match
+import ch.epfl.sdp.mobile.ui.social.ChessMatch
 
 /**
  * Composable that composes the PlayScreen [TODO] Contains a new game button only, should be
  * expanded to include history of matches
- * @param state state of the PlayScreen
+ * @param state the [PlayScreenState] to manage composable content
+ * @param key a function which uniquely identifies the list items.
  * @param modifier the [Modifier] for this composable.
```
alexandrepiveteau, line 27: Missing type parameter `M` documentation

mobile/src/main/java/ch/epfl/sdp/mobile/ui/play/PlayScreen.kt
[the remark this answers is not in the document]
alexandrepiveteau: I don't think items should be centered horizontally. I think the default arrangement and alignments are what we want.

mobile/src/main/java/ch/epfl/sdp/mobile/ui/profile/UserScreen.kt
```diff
@@ -22,6 +17,7 @@ import ch.epfl.sdp.mobile.ui.social.*
  * @param profileBar part of slot construct that represents the tab bar.
  * @param matches the part that is responsible for the list of all matches
  * @param lazyColumnState to keep state of LazyColumn
+ *
```
alexandrepiveteau, line 20: I think you meant to add some doc here but forgot ^^

```suggestion
 * onMatchClick a callback called when a [ChessMatch] is clicked
```

matt989253 (approved):
LGTM!

alexandrepiveteau (approved):

PR #214, opened 2022-04-03
Improve the following section UI

alexandrepiveteau:


mobile/src/main/java/ch/epfl/sdp/mobile/state/StatefulFollowingScreen.kt
```diff
@@ -114,8 +148,23 @@ private class SnapshotSocialScreenState(
...
+}
 
-  override fun onPersonClick(person: ProfileAdapter) {
-    onPersonItemClick(person)
+/**
+ * Reduces a [Flow] of [Interaction] to determine whether the flow represents a currently focused
+ * element. This is a [Flow]-based equivalent to [MutableInteractionSource]'s
+ * `collectIsFocusedAsState()`.
+ *
+ * @return the [Flow] of [Interaction] that is reduced.
```
BadrTad, line 158: [Question] I don't understand this doc. The fun returns a Flow that emits true iff the state is focused by reducing the interactions but it doesn't return the reduced flow itself as opposed to what the doc says. Or am I missing something ?

mobile/src/main/java/ch/epfl/sdp/mobile/ui/social/SocialScreen.kt
```diff
@@ -28,97 +33,135 @@ import ch.epfl.sdp.mobile.ui.social.SocialScreenState.Mode.Searching
...
+          style = MaterialTheme.typography.h4,
+          modifier = Modifier.padding(horizontal = 16.dp),
+      )
+    }
+    items(
+        items = players,
+        key = key,
+    ) { friend ->
+      PersonItem(
+          modifier = Modifier.clickable { onShowProfileClick(friend) }.animateItemPlacement(),
```
BadrTad, line 151: [Question] Does changing the order of `clickable` and `animateItemPlacement`  on the `Modifier ` results in a different behavior ?

BadrTad (dismissed):
Great work

mobile/src/main/java/ch/epfl/sdp/mobile/state/StatefulFollowingScreen.kt
alexandrepiveteau: Okay, apparently I had written `@return` instead of `@receiver` in the docs. This is fixed in 8d1c92e2a3b004f0bd461ea49dcd1025fec7d290.

mobile/src/main/java/ch/epfl/sdp/mobile/ui/social/SocialScreen.kt
alexandrepiveteau: Not in this case. In general, the ordering of `Modifier` may affect the behaviour of the items, however in this case `animateItemPlacement` is implemented as a `ParentDataModifier` so it will not be affected by `clickable {}` (or vice-versa).

`ParentDataModifier` is a way for layout nodes to pass some data to their parent, indicating that they want the parent layout to give them a specific behaviour (in this case, the children nodes want to tell their parent that they expect to be animated) [(source)](https://developer.android.com/reference/kotlin/androidx/compose/ui/layout/ParentDataModifier).

BadrTad (approved):

mobile/src/main/java/ch/epfl/sdp/mobile/ui/social/SocialScreen.kt
```diff
@@ -28,97 +33,135 @@ import ch.epfl.sdp.mobile.ui.social.SocialScreenState.Mode.Searching
...
 @OptIn(ExperimentalAnimationApi::class)
 @Composable
-fun <P : Person> SocialScreen(state: SocialScreenState<P>, modifier: Modifier = Modifier) {
-
-  val strings = LocalLocalizedStrings.current
-
+fun <P : Person> SocialScreen(
+    state: SocialScreenState<P>,
+    modifier: Modifier = Modifier,
+    key: ((P) -> Any)? = null,
```
Fouad-sys, line 44: [Question] Why use a function to uniquely identify the list items?

mobile/src/main/java/ch/epfl/sdp/mobile/state/StatefulFollowingScreen.kt
```diff
@@ -114,8 +148,23 @@ private class SnapshotSocialScreenState(
       }
     }
   }
+}
 
-  override fun onPersonClick(person: ProfileAdapter) {
-    onPersonItemClick(person)
+/**
+ * Reduces a [Flow] of [Interaction] to determine whether the flow represents a currently focused
```
Fouad-sys, line 154: [Question] What is meant by the word "focused"?

Fouad-sys (approved):

mobile/src/main/java/ch/epfl/sdp/mobile/ui/social/SocialScreen.kt
alexandrepiveteau: This is required to support animated items. Keys help Compose associate the state of each item with its associated composable : https://developer.android.com/jetpack/compose/lists#item-keys.

PR #222, opened 2022-04-05
Add an `id` accessor to `DocumentReference`

### LINKED ISSUE #220: Add a `document()` overload to the `Store` builder DSL

### LINKED ISSUE #221: Provide access to `DocumentReference.id` in `Store`

alexandrepiveteau:
This will close #221.

### DIFF
--- a/mobile/src/androidTest/java/ch/epfl/sdp/mobile/test/infrastructure/persistence/store/fake/FakeDocumentReference.kt
+++ b/mobile/src/androidTest/java/ch/epfl/sdp/mobile/test/infrastructure/persistence/store/fake/FakeDocumentReference.kt
@@ -23,6 +23,9 @@ class FakeDocumentReference(id: FakeDocumentId) : DocumentReference, CollectionB
 
   val state = MutableStateFlow(State(id))
 
+  override val id: String
+    get() = state.value.id.value
+
   override fun collection(path: String): CollectionReference {
     return state.updateAndGetWithValue {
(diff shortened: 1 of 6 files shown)


matt989253 (approved):
Thanks, LGTM!

mobile/src/androidTest/java/ch/epfl/sdp/mobile/test/infrastructure/persistence/store/fake/FakeDocumentReference.kt
```diff
@@ -23,6 +23,9 @@ class FakeDocumentReference(id: FakeDocumentId) : DocumentReference, CollectionB
 
   val state = MutableStateFlow(State(id))
 
+  override val id: String
```
BadrTad, line 26: [Question] I was always confused when it comes to overriding in Kotlin. When do we explicit the get ?

BadrTad (approved):

mobile/src/androidTest/java/ch/epfl/sdp/mobile/test/infrastructure/persistence/store/fake/FakeDocumentReference.kt
alexandrepiveteau: It depends on the behaviour you want to give. Long story shot, `override val id = "something"` creates a field (and **stores the value**) in the object, so it's a property. `override val id get() = "something"` works like a function and if what's in the `get()` changes over time, `id` may return different values if it's called multiple times.

PR #235, opened 2022-04-09
Change drawing order of available actions on chessboard

### LINKED ISSUE #234: Some actions / dots on the chessboard are hidden and not visible
As pointed out my @matt989253, in certain situations, some actions are not visible because they're hidden behind the drawing of the piece. The drawing order of the actions should be updated so the actions are always drawn on top of the chessboard.

## Example

In this situation, the pawn should be able to take the queen, but the action is not visible because the queen is opaque and covers the whole dot.

telegram-cloud-photo-size-4-5945082959801464774-y

alexandrepiveteau:
This fixes #234 by adding a `drawBehind` parameter to the `Modifier.cells` modifier. When set to `true`, the modifier will behave exactly as before. When set to `false`, the content will be drawn **before** the `Modifier`.

**Tests** : manual visual testing (we don't have screenshot tests yet but this would have been a good candidate)

## Screenshot :

In this situation, the pawn's action to eat the queen is drawn in the right order.

telegram-cloud-photo-size-4-5947021128218294633-y

### DIFF
--- a/mobile/src/main/java/ch/epfl/sdp/mobile/ui/game/ChessBoardModifiers.kt
+++ b/mobile/src/main/java/ch/epfl/sdp/mobile/ui/game/ChessBoardModifiers.kt
@@ -7,6 +7,7 @@ import androidx.compose.runtime.getValue
 import androidx.compose.ui.Modifier
 import androidx.compose.ui.composed
 import androidx.compose.ui.draw.drawBehind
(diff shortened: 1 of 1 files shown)


matt989253 (approved):

barmettlerl (approved):
It looks good for me. Nothing to add

PR #250, opened 2022-04-13
Add `Board.toBoard()` function to flatten boards, and flatten the past board history

alexandrepiveteau:
This change dramatically increases the performance of the chess engine, since it ensures that the boards are no longer wrapped in decorators when used in the next game step.

Combined with #236, this makes the game playable when 100+ moves are performed.

### DIFF
--- a/mobile/src/main/java/ch/epfl/sdp/mobile/application/chess/engine/Board.kt
+++ b/mobile/src/main/java/ch/epfl/sdp/mobile/application/chess/engine/Board.kt
@@ -1,5 +1,7 @@
 package ch.epfl.sdp.mobile.application.chess.engine
 
+import ch.epfl.sdp.mobile.application.chess.engine.implementation.buildBoard
+
 /**
  * A [Board] contains a bunch of [Piece] positioned in a certain way.
  *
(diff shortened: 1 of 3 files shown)


mobile/src/main/java/ch/epfl/sdp/mobile/application/chess/engine/Board.kt
```diff
@@ -30,3 +32,16 @@ interface Board<Piece : Any> : Iterable<Pair<Position, Piece>> {
...
+
+/**
+ * Copies the pieces from this [Board] into a new [Board]. The resulting [Board] is guaranteed to be
+ * immutable, and is created by copying all the pieces from the previous board.
+ *
+ * @param Piece the type of the pieces which are present in a board.
+ * @return a new [Board] copy.
+ */
+fun <Piece : Any> Board<Piece>.toBoard(): Board<Piece> = buildBoard {
+  for ((position, piece) in this@toBoard) {
```
barmettlerl, line 44: Question: What is the advantage to have a immutable Board. This copying is very expensive, depending on the game length.

barmettlerl (approved):
Code looks good. Have just a p'tit question :)

mobile/src/main/java/ch/epfl/sdp/mobile/application/chess/engine/Board.kt
alexandrepiveteau: An immutable `Board` is a good fit to model the current chess board state and make available (as separate `Board` instances) the past board states. The past board states are necessary to calculate some of the available moves (like whether castling is allowed or not).

However, I don't understand what you mean when saying "The copying is very expensive, depending on the game length". `toBoard()` works on a `Board`, not a `Game`, so it only creates a copy with the current pieces (aka at most 32 items). Do you have something else in mind ?

barmettlerl: Ah okey, mais cette restriction applies only to some pieces. Doesn't it make more sense to have a flag in the Board class that keeps track of those conditions?

KurohanaJuri (approved):

PR #253, opened 2022-04-14
Use different dispatchers when serializing and deserializing games

### LINKED ISSUE #237: Investigate and resolve lag issues in chess engine for games that have many moves
After some initial investigations, there seems to be two main issues which may decrease the performance of the chess board :

- When a new game state is received, the whole game has to get replayed. For long games, we'd like to see incremental updates applied.
- The color normalisation process does not flatten the boards regularly, resulting in game board accesses which have an increasing cost as the game moves on.

alexandrepiveteau:
This PR adds some basic multi-threading support to make moving pieces smoother.

This fixes #237, since the other points were addressed in previous PRs.


matt989253 (dismissed):

mobile/src/main/java/ch/epfl/sdp/mobile/application/chess/ChessFacade.kt
```diff
@@ -102,7 +113,8 @@ private data class StoreMatch(
 
   private val documentFlow = store.collection("games").document(id).asFlow<ChessDocument>()
 
-  override val game = documentFlow.map { it?.moves ?: emptyList() }.mapToGame()
+  override val game =
+      documentFlow.map { it?.moves ?: emptyList() }.mapToGame().buffer().flowOn(ioDispatcher)
```
alexandrepiveteau, line 117:
```suggestion
      documentFlow.map { it?.moves ?: emptyList() }.mapToGame().flowOn(ioDispatcher)
```

PR #255, opened 2022-04-14
add Profile name edit popup

barmettlerl:


mobile/src/androidTest/java/ch/epfl/sdp/mobile/test/state/StatefulHomeTest.kt
```diff
@@ -286,4 +286,70 @@ class StatefulHomeTest {
     rule.onNodeWithText("test2").assertExists()
     rule.onNodeWithText("test").assertExists()
   }
+
+  @Test
+  fun given_userIsLoggedIn_when_editProfileName_then_nameShouldBeUpdated() = runTest {
+    val auth = buildAuth { user("email@example.org", "password", "1") }
+    val store = buildStore {
+      collection("users") { document("1", ProfileDocument("1", name = "test", emoji = ":)")) }
```
alexandrepiveteau, line 294: You don't need to specify the document id here.

```suggestion
      collection("users") { document("1", ProfileDocument(name = "test", emoji = ":)")) }
```

mobile/src/androidTest/java/ch/epfl/sdp/mobile/test/state/StatefulHomeTest.kt
```diff
@@ -286,4 +286,70 @@ class StatefulHomeTest {
...
+
+    val strings =
+        rule.setContentWithLocalizedStrings {
+          val controller = rememberNavController()
+          ProvideFacades(authFacade, socialFacade, chessFacade) {
+            StatefulHome(
+                user = user,
+                controller = controller,
+            )
+          }
```
alexandrepiveteau, line 312: You should not have to hoist the `NavController` in this test.

```suggestion
          ProvideFacades(authFacade, socialFacade, chessFacade) {
            StatefulHome(
                user = user,
            )
          }
```

mobile/src/androidTest/java/ch/epfl/sdp/mobile/test/state/StatefulHomeTest.kt
```diff
@@ -286,4 +286,70 @@ class StatefulHomeTest {
...
+        }
+
+    rule.onNodeWithText(strings.sectionSettings).performClick()
+    rule.onNodeWithTag("editProfileName").assertExists()
+    rule.onNodeWithTag("editProfileName").performClick()
+    rule.onAllNodesWithText("test").assertCountEquals(2)
+    rule.onAllNodesWithText("test")[1].performTextInput("test2")
```
alexandrepiveteau, line 319: How can you be sure that the second node with text "test" is the right one ?

mobile/src/androidTest/java/ch/epfl/sdp/mobile/test/state/StatefulHomeTest.kt
```diff
@@ -286,4 +286,70 @@ class StatefulHomeTest {
...
+    rule.onNodeWithText(strings.settingEditSave).performClick()
+    rule.onAllNodesWithText("test").assertCountEquals(1)
```
alexandrepiveteau, line 321: I'm not sure I understand this. Don't we want to check that the username is indeed updated ?

mobile/src/androidTest/java/ch/epfl/sdp/mobile/test/state/StatefulHomeTest.kt
```diff
@@ -286,4 +286,70 @@ class StatefulHomeTest {
...
+                controller = controller,
+            )
+          }
+        }
+
+    rule.onNodeWithText(strings.sectionSettings).performClick()
+    rule.onNodeWithTag("editProfileName").assertExists()
+    rule.onNodeWithTag("editProfileName").performClick()
+    rule.onNodeWithText(strings.settingEditCancle).performClick()
+    rule.onAllNodesWithText("test").assertCountEquals(1)
```
alexandrepiveteau, line 353: If the count is expected to be 1, why not use `rule.onNodeWithText("test").assertIsDisplayed()` ?

mobile/src/androidTest/java/ch/epfl/sdp/mobile/test/state/StatefulHomeTest.kt
```diff
@@ -286,4 +286,70 @@ class StatefulHomeTest {
...
+            StatefulHome(
+                user = user,
+                controller = controller,
+            )
+          }
+        }
+
+    rule.onNodeWithText(strings.sectionSettings).performClick()
+    rule.onNodeWithTag("editProfileName").assertExists()
+    rule.onNodeWithTag("editProfileName").performClick()
```
alexandrepiveteau, line 351: You should use content descriptions, not test tags.

mobile/src/androidTest/java/ch/epfl/sdp/mobile/test/state/StatefulHomeTest.kt
```diff
@@ -286,4 +286,70 @@ class StatefulHomeTest {
...
+
+    val strings =
+        rule.setContentWithLocalizedStrings {
+          val controller = rememberNavController()
+          ProvideFacades(authFacade, socialFacade, chessFacade) {
+            StatefulHome(
+                user = user,
+                controller = controller,
+            )
+          }
```
alexandrepiveteau, line 346: Same as before, the `NavController` is not necessary based on my understanding.

```suggestion
          ProvideFacades(authFacade, socialFacade, chessFacade) {
            StatefulHome(
                user = user,
            )
          }
```

mobile/src/androidTest/java/ch/epfl/sdp/mobile/test/state/StatefulSettingScreenTest.kt
```diff
@@ -0,0 +1,37 @@
...
+      runTest {
+    val user = mockk<AuthenticatedUser>()
+    every { user.name } returns "test"
+    every { user.email } returns "test"
+    every { user.emoji } returns "test"
+    every { user.uid } returns "test"
+    every { user.followed } returns false
+    var state = false
+
+    val mockFunction = { state = true }
```
alexandrepiveteau, line 29: You may want to give `mockFunction` a different name or specify the arguments of `StatefulSettingsScreen` with named args; it's not really clear how the test works when reading it.

mobile/src/androidTest/java/ch/epfl/sdp/mobile/test/state/StatefulSettingScreenTest.kt
```diff
@@ -0,0 +1,37 @@
...
+
+    rule.setContent { StatefulSettingsScreen(user, mockFunction) }
+
+    rule.onNodeWithTag("editProfileName").performClick()
```
alexandrepiveteau, line 33: You should use content descriptions, not test tags.

mobile/src/androidTest/java/ch/epfl/sdp/mobile/test/ui/setting/SettingScreenTest.kt
```diff
@@ -26,8 +26,9 @@ class SettingScreenTest {
     override val pastGamesCount = 10
     override val puzzlesCount = 12
 
-    override fun onEditClick() = Unit
     override fun onSettingsClick() = Unit
+    override fun openEditProfileName() = Unit
```
alexandrepiveteau, line 30: You should stick to our existing convention of naming the callbacks `onBlahBlah`. `onEditProfileNameClick` may be an appropriate name for example.

mobile/src/main/java/ch/epfl/sdp/mobile/state/StatefulEditProfileDialog.kt
```diff
@@ -0,0 +1,62 @@
...
+ * @param user the current [AuthenticatedUser].
+ * @param scope the coroutine scope.
+ * @param onSave the callback called after we saved the changes.
+ * @param onCancle the callback called when we click the cancle button.
+ */
+class EditProfileNameDialogStateImpl(
+    user: AuthenticatedUser,
+    private val scope: CoroutineScope,
+    onSaveAction: State<() -> Unit>,
+    onCancleAction: State<() -> Unit>
```
alexandrepiveteau, line 22: Are these really different ? Shouldn't we only have an `closeDialog: State<() -> Unit>`-like parameter instead ?

mobile/src/main/java/ch/epfl/sdp/mobile/state/StatefulEditProfileDialog.kt
```diff
@@ -0,0 +1,62 @@
...
+) : EditProfileNameDialogState {
+  var user by mutableStateOf(user)
+  val onSaveAction by onSaveAction
+  val onCancleAction by onCancleAction
```
alexandrepiveteau, line 26: Should be `private`.

mobile/src/main/java/ch/epfl/sdp/mobile/state/StatefulEditProfileDialog.kt
```diff
@@ -0,0 +1,62 @@
...
+
+  override var name by mutableStateOf(user.name)
+
+  override fun onSaveClick() {
+    scope.launch {
+      user.update { name(name) }
+      onSaveAction()
```
alexandrepiveteau, line 33: You'll want to wrap this in a `MutatorMutex` lock, to handle the case where the user presses the save action two times very fast.

mobile/src/main/java/ch/epfl/sdp/mobile/state/StatefulEditProfileDialog.kt
```diff
@@ -0,0 +1,62 @@
...
+/**
+ * A stateful implementation of [EditProfileNameDialog] which uses some composition-local values to
+ * retrieve the appropriate dependencies.
+ *
+ * @param user the current [AuthenticatedUser].
+ * @param onSave the callback called after we saved the changes.
+ * @param onCancle the callback called when we click the cancle button.
+ */
+@Composable
+fun StatefulEditProfileDialog(user: AuthenticatedUser, onSave: () -> Unit, onCancle: () -> Unit) {
```
alexandrepiveteau, line 51: This is where you'll want to provide a single `onClose` action.

mobile/src/main/java/ch/epfl/sdp/mobile/state/StatefulHome.kt
```diff
@@ -81,7 +84,18 @@ fun StatefulHome(
...
+        StatefulSettingsScreen(
+            user = user,
+            modifier = Modifier.fillMaxSize(),
+            openEditProfileName = { controller.navigate(SettingEditProfileNameRoute) })
+      }
+      dialog(SettingEditProfileNameRoute) {
+        StatefulEditProfileDialog(
+            user = user,
+            onSave = { controller.navigate(SettingsRoute) },
+            onCancle = { controller.navigate(SettingsRoute) })
```
alexandrepiveteau, line 97: Maybe use `controller.popBackStack()` instead, which may handle pressing the back button better since it removes the dialog destination from the back stack.

mobile/src/main/java/ch/epfl/sdp/mobile/state/StatefulHome.kt
```diff
@@ -17,6 +17,9 @@ private const val SocialRoute = "social"
 /** The route associated to the settings tab. */
 private const val SettingsRoute = "settings"
 
+/** The route associated to editing button on the setting screen. */
+private const val SettingEditProfileNameRoute = "/profile-name/edit"
```
alexandrepiveteau, line 21: I'd try to keep the path consistent with the other routes.

```suggestion
private const val SettingEditProfileNameRoute = "profile-name/edit"
```

mobile/src/main/java/ch/epfl/sdp/mobile/state/StatefulSettingsScreen.kt
```diff
@@ -19,22 +18,29 @@ class AuthenticatedUserProfileScreenState(private val user: AuthenticatedUser) :
...
  * A stateful composable to visit setting page of the loged-in user
  *
  * @param user the current logged-in user.
+ * @param openEditProfileName Callable lambda to navigate to the profile Edit popup
  * @param modifier the [Modifier] for this composable.
  */
 @Composable
 fun StatefulSettingsScreen(
     user: AuthenticatedUser,
+    openEditProfileName: () -> Unit,
```
alexandrepiveteau, line 39: Naming inconsistent with the `onBlahBlah` convention.

mobile/src/main/java/ch/epfl/sdp/mobile/ui/i18n/English.kt
```diff
@@ -48,6 +48,10 @@ object English : LocalizedStrings {
   override val profileYourTurn = "Your turn to play!"
   override val profileOthersTurn = "Their turn to play."
 
+  override val settingEditSave = "Save".uppercase()
+  override val settingEditCancle = "Cancle".uppercase()
+  override val settingProfileName = "Profile Name"
```
alexandrepiveteau, line 53: Maybe `settingProfileNameTitle` ?

mobile/src/main/java/ch/epfl/sdp/mobile/ui/setting/SettingScreenState.kt
```diff
@@ -22,6 +22,6 @@ interface SettingScreenState : Person {
   /** On settings button clicked */
   fun onSettingsClick()
 
-  /** On edit button clicked */
-  fun onEditClick()
+  /** On edit profile name button clicked */
+  fun openEditProfileName()
```
alexandrepiveteau, line 26: Same comment on our existing naming convention.

mobile/src/main/java/ch/epfl/sdp/mobile/ui/setting/SettingScreen.kt
```diff
@@ -75,7 +77,14 @@ fun SettingHeader(state: SettingScreenState, modifier: Modifier = Modifier) {
   ) {
     SettingPicture(state)
     Column(horizontalAlignment = Alignment.CenterHorizontally) {
-      Text(state.name, style = MaterialTheme.typography.h5)
+      Row(horizontalArrangement = Arrangement.Center) {
+        Text(state.name, style = MaterialTheme.typography.h5)
+        IconButton(
+            onClick = state::openEditProfileName,
+            modifier = Modifier.size(35.dp).testTag("editProfileName")) {
```
alexandrepiveteau, line 84: Material Design is aligned on an 8 Dp grid. Maybe try with `24.dp` instead ? It's the standard icon size.

Also, you should probably apply the size modifier on the icon, not the wrapper `IconButton`.

alexandrepiveteau, line 84: Please don't use a `testTag` and prefer using the icon's content description (in this case, `strings.profileEditIcon`) in the tests.

mobile/src/main/java/ch/epfl/sdp/mobile/ui/setting/SettingScreen.kt
```diff
@@ -75,7 +77,14 @@ fun SettingHeader(state: SettingScreenState, modifier: Modifier = Modifier) {
   ) {
     SettingPicture(state)
     Column(horizontalAlignment = Alignment.CenterHorizontally) {
-      Text(state.name, style = MaterialTheme.typography.h5)
+      Row(horizontalArrangement = Arrangement.Center) {
```
alexandrepiveteau, line 80: You may want to align the content vertically using `CenterVertically`.

mobile/src/main/java/ch/epfl/sdp/mobile/ui/setting/EditProfileNameDialog.kt
```diff
@@ -0,0 +1,44 @@
...
+ * @param state the [EditProfileNameDialogState] as an argument.
+ * @param modifier the [Modifier] for this composable.
+ */
+@Composable
+fun EditProfileNameDialog(
+    state: EditProfileNameDialogState,
+    modifier: Modifier = Modifier,
+) {
+
+  Box(modifier.fillMaxSize(), Alignment.Center) {
```
alexandrepiveteau, line 26: Could you please add a comment regarding the reason for which there's this `Box` ?

mobile/src/main/java/ch/epfl/sdp/mobile/ui/setting/EditProfileNameDialog.kt
```diff
@@ -0,0 +1,44 @@
...
+    val strings = LocalLocalizedStrings.current
+
+    Dialog(
+        modifier = modifier,
```
alexandrepiveteau, line 30: You must not use the `modifier` outside of the top-level composable.

```suggestion
```

mobile/src/main/java/ch/epfl/sdp/mobile/ui/setting/EditProfileNameDialog.kt
```diff
@@ -0,0 +1,44 @@
...
+import androidx.compose.material.TextField
+import androidx.compose.runtime.Composable
+import androidx.compose.ui.Alignment
+import androidx.compose.ui.Modifier
+import androidx.compose.ui.unit.dp
+import ch.epfl.sdp.mobile.state.LocalLocalizedStrings
+import ch.epfl.sdp.mobile.ui.prepare_game.Dialog
+
+/**
+ * Component for the Setting profile screen
```
alexandrepiveteau, line 15: You may want to specify that this is a dialog that lets the user edit their profile name ^^

mobile/src/main/java/ch/epfl/sdp/mobile/ui/setting/EditProfileNameDialog.kt
```diff
@@ -0,0 +1,44 @@
...
+
+    Dialog(
+        modifier = modifier,
+        cancelContent = { Text(text = strings.settingEditCancle) },
+        confirmContent = { Text(text = strings.settingEditSave) },
+        onCancelClick = state::onCancleClick,
+        onConfirmClick = state::onSaveClick) {
+      Column(modifier = modifier.padding(16.dp)) {
+        Text(strings.settingProfileName, style = MaterialTheme.typography.subtitle1)
+        TextField(
```
alexandrepiveteau, line 37: Is it expected that there is no spacing between the items ?

alexandrepiveteau (commented):
Nice work ! It looks good overall, however a few changes are required to the tests and some minor adjustments are needed to the implementation. Thanks for what you've already done !

Also, an important detail : you've written **cancle** instead of **cancel** in many places, so I suggest you Ctrl+R it everywhere ^^

mobile/src/main/java/ch/epfl/sdp/mobile/ui/setting/EditProfileNameDialog.kt
barmettlerl: Ah sure, is done

mobile/src/main/java/ch/epfl/sdp/mobile/ui/setting/EditProfileNameDialog.kt
barmettlerl: I'll add it

mobile/src/main/java/ch/epfl/sdp/mobile/ui/setting/SettingScreen.kt
barmettlerl: Good point

mobile/src/androidTest/java/ch/epfl/sdp/mobile/test/state/StatefulSettingScreenTest.kt
barmettlerl: How can I use content description to pin down an Icon ?

mobile/src/androidTest/java/ch/epfl/sdp/mobile/test/ui/setting/SettingScreenTest.kt
barmettlerl: Ah there is a convention. Good to know -> is changed

mobile/src/main/java/ch/epfl/sdp/mobile/state/StatefulEditProfileDialog.kt
barmettlerl: No they are exactly the same. However, I want to make the code more readable and thus I separated them.

barmettlerl: So it is similar to the other dialog we have in `mobile/src/main/java/ch/epfl/sdp/mobile/ui/prepare_game/PrepareGameScreenState.kt`

mobile/src/main/java/ch/epfl/sdp/mobile/state/StatefulEditProfileDialog.kt
barmettlerl: Yes, you're right

mobile/src/main/java/ch/epfl/sdp/mobile/state/StatefulEditProfileDialog.kt
barmettlerl: As explained previously, I want to make the interface to the Dialog similar to the other dialog we have.

mobile/src/main/java/ch/epfl/sdp/mobile/state/StatefulHome.kt
barmettlerl: 👌 is changed

mobile/src/main/java/ch/epfl/sdp/mobile/ui/setting/EditProfileNameDialog.kt
barmettlerl: Ah, true -> is changed

mobile/src/androidTest/java/ch/epfl/sdp/mobile/test/state/StatefulSettingScreenTest.kt
alexandrepiveteau: Something along the lines of :

```
val strings = rule.setContentWithLocalizedStrings { Icon(..., contentDescription = LocalLocalizedStrings.current.something) )}
rule.onNodeWithContentDescription(strings.something).performClick()
```

mobile/src/main/java/ch/epfl/sdp/mobile/state/StatefulEditProfileDialog.kt
```diff
@@ -0,0 +1,70 @@
...
+  /**
+   * A [MutatorMutex] which enforces mutual exclusion of update profile name requests. Performing a
+   * new request (by clicking the button) will cancel the currently pending request.
+   */
+  private val mutex = MutatorMutex()
+
+  override fun onSaveClick() {
+    scope.launch {
+      user.update { name(name) }
+      mutex.mutate(MutatePriority.UserInput) { onSaveAction() }
```
alexandrepiveteau, line 41: You should wrap the `user.update` in the `mutate` block too.

alexandrepiveteau: So something like :

```kotlin
mutex.mutate(MutatePriority.UserInput) {
  user.update { name(name) }
  onSaveAction()
}
```

mobile/src/main/java/ch/epfl/sdp/mobile/state/StatefulSettingsScreen.kt
```diff
@@ -19,22 +20,29 @@ class AuthenticatedUserProfileScreenState(private val user: AuthenticatedUser) :
   override val name = user.name
   override val emoji = user.emoji
   override val followed = user.followed
+  val onEditProfileNameClickAction by onEditProfileNameClickAction
```
alexandrepiveteau, line 23: Visibility isn't appropriate imho.

```suggestion
  private val onEditProfileNameClickAction by onEditProfileNameClickAction
```

mobile/src/main/java/ch/epfl/sdp/mobile/state/StatefulSettingsScreen.kt
```diff
@@ -1,16 +1,17 @@
...
 import androidx.compose.ui.Modifier
 import ch.epfl.sdp.mobile.application.Profile.Color
 import ch.epfl.sdp.mobile.application.authentication.AuthenticatedUser
 import ch.epfl.sdp.mobile.ui.setting.SettingScreenState
 import ch.epfl.sdp.mobile.ui.setting.SettingsScreen
 import ch.epfl.sdp.mobile.ui.social.ChessMatch
 
-class AuthenticatedUserProfileScreenState(private val user: AuthenticatedUser) :
-    SettingScreenState {
+class SettingsScreenState(
```
alexandrepiveteau, line 11: It's strange having `SettingsScreenState` which implements `SettingsScreenState`. What about naming the implementation `SnapshotSettingsScreenState` ?

mobile/src/main/java/ch/epfl/sdp/mobile/state/StatefulSettingsScreen.kt
```diff
@@ -19,22 +20,29 @@ class AuthenticatedUserProfileScreenState(private val user: AuthenticatedUser) :
...
+  override fun onEditProfileNameClick() {
+    onEditProfileNameClickAction()
+  }
 }
 
 /**
  * A stateful composable to visit setting page of the loged-in user
  *
  * @param user the current logged-in user.
+ * @param openEditProfileName Callable lambda to navigate to the profile Edit popup
```
alexandrepiveteau, line 35: Please update the documentation.

mobile/src/main/java/ch/epfl/sdp/mobile/ui/setting/EditProfileNameDialogState.kt
```diff
@@ -0,0 +1,12 @@
+package ch.epfl.sdp.mobile.ui.setting
+
+/** State interface of the [EditProfileNameDialog] */
+interface EditProfileNameDialogState {
+  var name: String
```
alexandrepiveteau, line 5: Missing documentation.

mobile/src/main/java/ch/epfl/sdp/mobile/ui/setting/EditProfileNameDialog.kt
```diff
@@ -0,0 +1,52 @@
...
+  */
+  Box(modifier.fillMaxSize(), Alignment.Center) {
+    val strings = LocalLocalizedStrings.current
+
+    Dialog(
+        cancelContent = { Text(text = strings.settingEditCancel) },
+        confirmContent = { Text(text = strings.settingEditSave) },
+        onCancelClick = state::onCancelClick,
+        onConfirmClick = state::onSaveClick) {
+      Column(modifier = modifier.padding(16.dp), verticalArrangement = Arrangement.spacedBy(8.dp)) {
```
alexandrepiveteau, line 42: Bug with improper modifier being used.

```suggestion
      Column(modifier = Modifier.padding(16.dp), verticalArrangement = Arrangement.spacedBy(8.dp)) {
```

mobile/src/main/java/ch/epfl/sdp/mobile/state/StatefulEditProfileDialog.kt
```diff
@@ -0,0 +1,70 @@
...
+/**
+ * A stateful implementation of [EditProfileNameDialog] which uses some composition-local values to
+ * retrieve the appropriate dependencies.
+ *
+ * @param user the current [AuthenticatedUser].
+ * @param onSave the callback called after we saved the changes.
+ * @param onCancel the callback called when we click the cancel button.
+ */
+@Composable
+fun StatefulEditProfileDialog(user: AuthenticatedUser, onSave: () -> Unit, onCancel: () -> Unit) {
```
alexandrepiveteau, line 59: I think `StatefulEditProfileNameDialog` is a better name, and the file name should be updated too.

alexandrepiveteau (commented):
What makes me perplex with the two arguments for the `StatefulEditProfileDialog` is that `onSave` and `onCancel` don't have a real meaning. The stateful composable wants to know how to navigate when the dialog is dismissed, but that's pretty much it.

This is different from the `PrepareGameDialog`, which acts differently when a game is created (it must navigate to the new game) or when the dialog is dismissed (it must navigate back). In that case, the destinations will be different, but this is not true for the `StatefulEditProfileDialog`.

Therefore, I'd strongly advise to only pass a single parameter to the dialog that indicates how to dismiss it.

barmettlerl:
> What makes me perplex with the two arguments for the `StatefulEditProfileDialog` is that `onSave` and `onCancel` don't have a real meaning. The stateful composable wants to know how to navigate when the dialog is dismissed, but that's pretty much it.
> 
> This is different from the `PrepareGameDialog`, which acts differently when a game is created (it must navigate to the new game) or when the dialog is dismissed (it must navigate back). In that case, the destinations will be different, but this is not true for the `StatefulEditProfileDialog`.
> 
> Therefore, I'd strongly advise to only pass a single parameter to the dialog that indicates how to dismiss it.

Thank you for your detailed description. However, I don't think this makes a big difference and it increases the readability which is much more important (in my opinion) than logical correctness. So I will keep it as it is.

alexandrepiveteau:
> Thank you for your detailed description. However, I don't think this makes a big difference and it increases the readability which is much more important (in my opinion) than logical correctness. So I will keep it as it is. 

I don't quite understand how code duplication will help with readability, especially in this specific case (since we could reduce the cognitive complexity of the code by removing a duplicated argument). However, we'll be introducing some small technical debt which could be easily addressed now.

May I suggest asking some other team members on whether they think it increases readability for them before making a decision on this point ?

mobile/src/main/java/ch/epfl/sdp/mobile/state/StatefulEditProfileNameDialog.kt
```diff
@@ -0,0 +1,75 @@
...
+import kotlinx.coroutines.CoroutineScope
+import kotlinx.coroutines.launch
+
+/**
+ * Implementation of the [EditProfileNameDialogState] interface
+ *
+ * @param user the current [AuthenticatedUser].
+ * @param scope the coroutine scope.
+ * @param onSave the callback called after we saved the changes.
+ * @param onCancel the callback called when we click the cancel button.
```
Fouad-sys, line 18: Parameter names should be updated.
```suggestion
 * @param onSaveAction the callback called after we saved the changes.
 * @param onCancelAction the callback called when we click the cancel button.
```

mobile/src/main/java/ch/epfl/sdp/mobile/state/StatefulHome.kt
```diff
@@ -17,6 +17,9 @@ private const val SocialRoute = "social"
 /** The route associated to the settings tab. */
 private const val SettingsRoute = "settings"
 
+/** The route associated to editing button on the setting screen. */
```
Fouad-sys, line 20:
```suggestion
/** The route associated to the editing button on the setting screen. */
```

mobile/src/main/java/ch/epfl/sdp/mobile/state/StatefulEditProfileNameDialog.kt
```diff
@@ -0,0 +1,75 @@
...
+  /**
+   * A [MutatorMutex] which enforces mutual exclusion of update profile name requests. Performing a
+   * new request (by clicking the button) will cancel the currently pending request.
+   */
+  private val mutex = MutatorMutex()
+
+  override fun onSaveClick() {
+    scope.launch {
+      mutex.mutate(MutatePriority.UserInput) {
+        user.update { name(name) }
```
Fouad-sys, line 40: [Question] Does name(name) change the user's name? If so maybe one of both could have a different name.

Fouad-sys (commented):
Looks great! Maybe a few other nitpicks to improve would be to darken the user's name color in the popup and add a line separating the profile name and cancel/save parts, just like in the figma prototypes. But these are not necessary to merge this PR and can be changed in the prototypes.

Fouad-sys:
> > Thank you for your detailed description. However, I don't think this makes a big difference and it increases the readability which is much more important (in my opinion) than logical correctness. So I will keep it as it is.
> 
> I don't quite understand how code duplication will help with readability, especially in this specific case (since we could reduce the cognitive complexity of the code by removing a duplicated argument). However, we'll be introducing some small technical debt which could be easily addressed now.
> 
> May I suggest asking some other team members on whether they think it increases readability for them before making a decision on this point ?

I agree that having two callback parameters that perform the same action could be avoided, but I do see Lars's view to have a callback for each button click, just for intuition, so I personally would be fine with both implementation choices.

mobile/src/main/java/ch/epfl/sdp/mobile/state/StatefulEditProfileNameDialog.kt
barmettlerl: Yes, `name(name)` sends a request on firebase to update the username.

Good input, it is not quite intuitive what this line of code does. I will rename the `user.name->name` to  `user.name->userName`.

barmettlerl:
> > Thank you for your detailed description. However, I don't think this makes a big difference and it increases the readability which is much more important (in my opinion) than logical correctness. So I will keep it as it is.
> 
> I don't quite understand how code duplication will help with readability, especially in this specific case (since we could reduce the cognitive complexity of the code by removing a duplicated argument). However, we'll be introducing some small technical debt which could be easily addressed now.
> 
> May I suggest asking some other team members on whether they think it increases readability for them before making a decision on this point ?



> > > Thank you for your detailed description. However, I don't think this makes a big difference and it increases the readability which is much more important (in my opinion) than logical correctness. So I will keep it as it is.
> > 
> > 
> > I don't quite understand how code duplication will help with readability, especially in this specific case (since we could reduce the cognitive complexity of the code by removing a duplicated argument). However, we'll be introducing some small technical debt which could be easily addressed now.
> > May I suggest asking some other team members on whether they think it increases readability for them before making a decision on this point ?
> 
> I agree that having two callback parameters that perform the same action could be avoided, but I do see Lars's view to have a callback for each button click, just for intuition, so I personally would be fine with both implementation choices.

Haha, your neutrality does not really resolve the conflict here.

mobile/src/androidTest/java/ch/epfl/sdp/mobile/test/state/StatefulHomeTest.kt
barmettlerl: Same as https://github.com/epfl-SDP/android/pull/255/files#r851789468

mobile/src/androidTest/java/ch/epfl/sdp/mobile/test/state/StatefulHomeTest.kt
barmettlerl: I've tried to implement it like this, however it seems that in the test environment the name is not updated on the `SettingsScreen`. So I just tested if the Dialog is closed. Do you know what could be the problem?

mobile/src/main/java/ch/epfl/sdp/mobile/application/chess/ChessFacade.kt
```diff
@@ -36,7 +36,7 @@ class ChessFacade(private val auth: Auth, private val store: Store) {
   }
 
   /**
-   * Returns the [Match] associated to the given identifier.
+   * <<<<<<< HEAD Returns the [Match] associated to the given identifier.
```
alexandrepiveteau, line 39: Looks like a VCS issue.

```suggestion
   * Returns the [Match] associated to the given identifier.
```

mobile/src/main/java/ch/epfl/sdp/mobile/application/chess/ChessFacade.kt
```diff
@@ -45,8 +45,8 @@ class ChessFacade(private val auth: Auth, private val store: Store) {
   }
 
   /**
-   * Fetches a [Flow] of [List] of [Match]s that a certain [Profile] has going on with any other
-   * player (or even himself)
+   * ======= >>>>>>> display-active-games Fetches a [Flow] of [List] of [Match]s that a certain
```
alexandrepiveteau, line 48: Looks like a VCS issue.

```suggestion
   * Fetches a [Flow] of [List] of [Match]s that a certain
```

mobile/src/androidTest/java/ch/epfl/sdp/mobile/test/state/StatefulSettingScreenTest.kt
alexandrepiveteau: Could you please copy the stack trace from the test here ?

There could be many causes for this problem. For instance, it's possible that multiple nodes match the content description if it's not specific enough for instance.

mobile/src/androidTest/java/ch/epfl/sdp/mobile/test/state/StatefulHomeTest.kt
alexandrepiveteau: I highly suggest you use a less-flaky way to retrieve the text input node. The `onAllNodesWithText` documentation does not give any guarantee on the ordering of the nodes matched in the semantics tree.

You could probably use a custom matcher, like this for instance :

```kotlin
rule.onNode(hasSetTextAction() and hasText("test")).performWhatever()
```

barmettlerl: Is done thx

mobile/src/androidTest/java/ch/epfl/sdp/mobile/test/state/StatefulHomeTest.kt
barmettlerl: Solved it

mobile/src/main/java/ch/epfl/sdp/mobile/state/StatefulSettingsScreen.kt
```diff
@@ -1,37 +1,49 @@
...
+    onEditProfileNameClickAction: State<() -> Unit>
 ) : SettingScreenState, Person by ProfileAdapter(user) {
   override val email = user.email
   override val pastGamesCount = 0
   override val puzzlesCount = 0
   override val matches = emptyList<ChessMatch>()
+  override val backgroundColor = Profile.Color.Orange.toColor()
+  override val name = user.name
+  override val emoji = user.emoji
+  override val followed = user.followed
```
alexandrepiveteau, line 23: These fields are implemented using `Person by ProfileAdapter(user)`.

```suggestion
```

alexandrepiveteau (approved):
LGTM, thanks !

alexandrepiveteau (dismissed):

mobile/src/main/java/ch/epfl/sdp/mobile/state/StatefulSettingsScreen.kt
```diff
@@ -1,37 +1,44 @@
...
 import androidx.compose.ui.Modifier
 import ch.epfl.sdp.mobile.application.authentication.AuthenticatedUser
 import ch.epfl.sdp.mobile.ui.setting.SettingScreenState
 import ch.epfl.sdp.mobile.ui.setting.SettingsScreen
 import ch.epfl.sdp.mobile.ui.social.ChessMatch
 import ch.epfl.sdp.mobile.ui.social.Person
 
-class AuthenticatedUserProfileScreenState(
-    private val user: AuthenticatedUser,
+class SettingsScreenStateImpl(
```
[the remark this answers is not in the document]
barmettlerl: thx, is done

alexandrepiveteau (approved):

KurohanaJuri (approved):

PR #256, opened 2022-04-18
Implement promotion UI and game engine support

alexandrepiveteau:
The declarative `Action` + `NextStep` mechanism of the chess engine is improved to add support for promotion. Promotion is implemented by providing multiple `Action` with the same start position and delta, but with different ranks associated.

The user interface is improved to let the user disambiguate which `Action` is picked when multiple actions are available from the same origin and destination.

---

## Detailed changes :

- `NextStep.MovePiece` now requires an `Action` to select the next action, instead of a `Position` and a `Delta`. This disambiguates promotion actions.
- `Action` is now a `sealed interface`, with the subclasses `Action.Move` and `Action.Promote`. `Action.Move` replaces the previous `Action` class, while `Promote` adds a `val rank: Rank` which is used for promotion actions.
- `PersistentPieceIdentifier` is removed, and `PieceIdentifier` now contains the unique disambiguator for pieces.
- `Effect.move` now takes two `Position` instead of a `Position` and a `Delta`.
- `Effect.promote` is added as an extension of `Effect.move` which replaces the target position with a piece.
- `delta` now accepts an additional `promotionAllowed` parameter.
- `Action` is now an interface with two implementations `Move` and `Promote`.
- The notation pipeline is completely overhauled with support for promotion notation, and uses parser combinators under-the-hood.
- A `PromoteDialog` composable is added, alongside a `PromotionState` interface.
- `SnapshotChessBoardState` now implements `PromotionState`.


mobile/src/main/java/ch/epfl/sdp/mobile/application/chess/engine/rules/Moves.kt
```diff
@@ -36,19 +39,51 @@ typealias Moves<P> = Sequence<Pair<Action, Effect<P>>>
...
+    promotionAllowed: Boolean,
+) {
+  val isOnLastRow = to.y == 0
+  if (!isOnLastRow || !promotionAllowed) {
+    yield(Move(from, to) to move(from, to))
+  } else {
+    // Retrieve the origin piece.
+    val piece = board[from] ?: return
+    // Compute the identifier of the promoted piece.
+    val id = board.maxOf { (_, piece) -> piece.id }.inc()
```
matt989253, line 82: Is there a reason you need to compute a new identifier for the promoted piece? Could you not just reuse the pawn's identifier, since it will be completely replaced?

mobile/src/main/java/ch/epfl/sdp/mobile/state/StatefulGameScreen.kt
```diff
@@ -68,8 +71,33 @@ fun StatefulGameScreen(
...
+ */
+@Composable
+private fun StatefulPromoteDialog(
+    state: PromotionState,
+    modifier: Modifier = Modifier,
+) {
+  if (state.choices.isNotEmpty()) {
+    PromoteDialog(
+        // Always use the Black color to remain consistent with Figma ?
+        color = ChessBoardState.Color.Black,
```
matt989253, line 87: Although it's clearly not a priority, I think it would be good to have a low-priority issue for displaying the correct color here.

alexandrepiveteau: Referenced in https://github.com/epfl-SDP/android/issues/265.

mobile/src/main/java/ch/epfl/sdp/mobile/application/chess/engine/rules/Moves.kt
alexandrepiveteau: Fixed in 0f09f58d9ca27f5e48bbe259823344e119396855.

mobile/src/main/java/ch/epfl/sdp/mobile/ui/game/PromoteDialog.kt
```diff
@@ -0,0 +1,138 @@
...
+) {
+  val strings = LocalLocalizedStrings.current
+  val tint by animateColorAsState(if (selected) PawniesColors.Orange200 else PawniesColors.Green800)
+  Icon(
+      painter = rank.icon(color = color),
+      contentDescription = rank.contentDescription(strings),
+      tint = tint,
+      modifier =
+          modifier
+              .clip(CircleShape)
```
barmettlerl, line 49: Why you use clip here ?

barmettlerl (approved):
It is a quite big PR. Maybe try next time to make smaller chunks, which are easier to review

mobile/src/main/java/ch/epfl/sdp/mobile/application/chess/engine/rules/Action.kt
```diff
@@ -2,13 +2,66 @@ package ch.epfl.sdp.mobile.application.chess.engine.rules
...
-data class Action(val from: Position, val delta: Delta)
+sealed interface Action {
+
+  /** The position from which the piece is dragged. */
+  val from: Position
+
+  /** The amount by which the piece is moved. */
+  val delta: Delta
+
+  // Convenience operators to keep a destructuring-friendly syntax.
```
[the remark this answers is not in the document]
alexandrepiveteau: `position` and `delta` are made available through component destructuring : https://kotlinlang.org/docs/destructuring-declarations.html. This way, one could write some code as follows :

```kotlin
val action: Action = // something ...
val (position, delta) = action
```

even if `action` is an interface and not a `data class`.

mobile/src/main/java/ch/epfl/sdp/mobile/application/chess/engine/PieceIdentifier.kt
```diff
@@ -5,4 +5,12 @@ package ch.epfl.sdp.mobile.application.chess.engine
  * between board positions, since pieces with the same stable ids can have their positions
  * interpolated.
  */
-interface PieceIdentifier
+data class PieceIdentifier(private val id: Int) : Comparable<PieceIdentifier> {
+
+  override operator fun compareTo(other: PieceIdentifier): Int = id.compareTo(other.id)
+
+  /** Returns an incremented [PieceIdentifier]. */
+  operator fun inc(): PieceIdentifier {
```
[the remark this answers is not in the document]
alexandrepiveteau: Both approaches are possible, however generating new `PieceIdentifier` using `inc()` has the advantage of not relying on the actual structure of the backing identifiers (in this case, an `Int`). This way, `id` can remain a `private val` in `PieceIdentifier`.

mobile/src/main/java/ch/epfl/sdp/mobile/application/chess/engine/rules/Effect.kt
```diff
@@ -58,19 +57,37 @@ fun interface Effect<Piece : Any> {
...
-     * @param delta the [Delta] applied to the piece. Must stay within the bounds.
+     * @param target the final position of the piece.
      * @return the valid resulting [Effect].
      */
-    fun <P : Any> move(from: Position, delta: Delta): Effect<P> = Effect {
+    fun <P : Any> move(from: Position, target: Position): Effect<P> = Effect {
       val original = it[from] ?: return@Effect it
-      val final = from + delta ?: return@Effect it
-      replace(final, original).perform(remove<P>(from).perform(it))
+      if (!target.inBounds) return@Effect it
```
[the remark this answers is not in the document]
alexandrepiveteau: This is a return with an implicit label https://kotlinlang.org/docs/returns.html#return-to-labels. It means that it returns from the lambda.

mobile/src/main/java/ch/epfl/sdp/mobile/ui/game/PromoteDialog.kt
alexandrepiveteau: To clip the ripple from the `clickable` modifier.

matt989253 (approved):

PR #257, opened 2022-04-20
Introduce `sharedTest` source set and move many UI tests to use Robolectric

alexandrepiveteau:
This PR introduces the following changes :

- A new `sharedTest` source set is introduced.
- The existing tests are moved to the `test` and `androidTest` subpackages.
- Many UI tests are now run as Unit tests using Robolectric, resulting in considerably faster builds.


alexandrepiveteau:
We will probably not be able to merge Robolectric support until https://github.com/robolectric/robolectric/issues/7055 is solved, since it makes some of our tests flaky and fail on the CI while they succeed locally.

PR #266, opened 2022-04-26
Remove the legacy `Match()` constructor

alexandrepiveteau:
This removes the `Match` builder which has become irrelevant since `StoreMatch` was introduced, and the method `update` was added to `Match`.


matt989253 (approved):

Fouad-sys (approved):
LGTM

PR #267, opened 2022-04-26
Fix a MVVM violation in `Person`

alexandrepiveteau:
The `backgroundColor` property of `Person` is a `ch.epfl.sdp.mobile.application.Profile.Color`, which is a model concept. This PR changes its type to `androidx.compose.ui.graphics.Color`, since `Person` is part of the view.

### DIFF
--- a/mobile/src/main/java/ch/epfl/sdp/mobile/state/StatefulSettingsScreen.kt
+++ b/mobile/src/main/java/ch/epfl/sdp/mobile/state/StatefulSettingsScreen.kt
@@ -3,22 +3,19 @@ package ch.epfl.sdp.mobile.state
 import androidx.compose.runtime.Composable
 import androidx.compose.runtime.remember
 import androidx.compose.ui.Modifier
-import ch.epfl.sdp.mobile.application.Profile.Color
 import ch.epfl.sdp.mobile.application.authentication.AuthenticatedUser
 import ch.epfl.sdp.mobile.ui.setting.SettingScreenState
 import ch.epfl.sdp.mobile.ui.setting.SettingsScreen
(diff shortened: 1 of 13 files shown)


mobile/src/main/java/ch/epfl/sdp/mobile/state/StatefulSettingsScreen.kt
```diff
@@ -3,22 +3,19 @@ package ch.epfl.sdp.mobile.state
...
 import ch.epfl.sdp.mobile.ui.setting.SettingScreenState
 import ch.epfl.sdp.mobile.ui.setting.SettingsScreen
 import ch.epfl.sdp.mobile.ui.social.ChessMatch
+import ch.epfl.sdp.mobile.ui.social.Person
 
-class AuthenticatedUserProfileScreenState(private val user: AuthenticatedUser) :
-    SettingScreenState {
+class AuthenticatedUserProfileScreenState(
+    private val user: AuthenticatedUser,
+) : SettingScreenState, Person by ProfileAdapter(user) {
```
Fouad-sys, line 14: [Question] Why make the AuthenticatedUserProfileScreenState a Person?

Fouad-sys (approved):
Good spot and application of separation of concerns, thanks!

mobile/src/main/java/ch/epfl/sdp/mobile/state/StatefulSettingsScreen.kt
alexandrepiveteau: It's already a `Person`, since `SettingScreenState : Person`.

PR #268, opened 2022-04-27
Speech recognizer output filtering

BadrTad:
This PR implements a filtering unit for the speech recognizer results. It relies on a dictionary of chess words and a set of rules to determine the spoken user chess command. It outputs the command with the format {ChessPiece#LetterNumber} e.g {Queen#B6}.
~~The filtering class also outputs partial results, for example if it has detected only the letter and the chess piece  but not the number then the output will be {ChessPiece#Letter$} e.g {Queen#B$}~~
The filtering class either outputs the full chess command or null


KurohanaJuri (commented):
I have a question about the dictionary. How did you find the list of homophone? You try the speech recognizer or it from an existing homophone dictionary ?

BadrTad:
I tried the created the inputs manually by recognizing my speech which could, indeed, be biased but it was just a first step to see if the filtering code works. I agree that we should be using some pre-esixting dictionary.

mobile/src/main/java/ch/epfl/sdp/mobile/ui/speech_recognition/ChessSpeechRecognitionFilter.kt
```diff
@@ -0,0 +1,210 @@
+package ch.epfl.sdp.mobile.ui.speech_recognition
+
+import java.lang.StringBuilder
```
KurohanaJuri, line 3: You don't need to import `StringBuilder`. Kotlin has this already integrated

mobile/src/main/java/ch/epfl/sdp/mobile/ui/speech_recognition/ChessSpeechRecognitionFilter.kt
```diff
@@ -0,0 +1,210 @@
...
+    for (candidate in candidates) {
+      if (candidate.value > bestCandidate.value) {
+        bestCandidate = candidate
+      }
+    }
+    return bestCandidate.key
+  }
+
+  /**
+   * Detects if any word in the speech is within the numberDictionary and return its equivalent
```
KurohanaJuri, line 76:
```suggestion
   * Detects if any word in the speech is within the pieceDictionary and return its equivalent
```

mobile/src/main/java/ch/epfl/sdp/mobile/ui/speech_recognition/ChessSpeechRecognitionFilter.kt
```diff
@@ -0,0 +1,210 @@
...
+   */
+  private fun detectPieceType(
+      words: List<Word>,
+      pieceDictionary: Map<ChessPiece, List<String>> = chessDictionary.chessPieces
+  ): String {
+    return detect(words, pieceDictionary)
+  }
+
+  /**
+   * Detects if any word in the speech is within the numberDictionary and return its equivalent
```
KurohanaJuri, line 90:
```suggestion
   * Detects if any word in the speech is within the lettersDictionary and return its equivalent
```

mobile/src/main/java/ch/epfl/sdp/mobile/ui/speech_recognition/ChessSpeechRecognitionFilter.kt
```diff
@@ -0,0 +1,210 @@
...
+   */
+  private fun detectNumber(
+      words: List<Word>,
+      numbersDictionary: Map<ChessNumber, List<String>> = chessDictionary.numbers
+  ): String {
+    return detect(words, numbersDictionary)
+  }
+
+  /**
+   * Makes a decision of choosing a conform by relying on confidence scores
```
KurohanaJuri, line 118: The documentation is not really clear, can you rewrite this to be more explicit ?

mobile/src/main/java/ch/epfl/sdp/mobile/ui/speech_recognition/ChessSpeechRecognitionFilter.kt
```diff
@@ -0,0 +1,210 @@
...
+      if (candidate.isNotEmpty()) {
+        candidates[candidate] = confidencesScores[index]
+      }
+    }
+
+    return bestCandidate(candidates)
+  }
+
+  /**
+   * Searches for any matches of for a list word given a dictionary of chess keywords
```
KurohanaJuri, line 143:
```suggestion
   * Searches for any matches for a list word given a dictionary of chess keywords
```

mobile/src/main/java/ch/epfl/sdp/mobile/ui/speech_recognition/ChessSpeechRecognitionFilter.kt
```diff
@@ -0,0 +1,210 @@
...
+                true -> pieceFromRule
+                else -> "!!"
+              }
+        }
+
+    // Apply filters for letter and number
+    val letter =
+        detectFromSpeeches(tokenizedSpeeches, confidencesScores, this::detectLetter).ifEmpty { "@" }
+    val number =
+        detectFromSpeeches(tokenizedSpeeches, confidencesScores, this::detectNumber).ifEmpty { "$" }
```
KurohanaJuri, line 195: I don't think it's a good idea to have these magic number to define that we don't have a correspondence.

I think that if one element didn't match, we just return an empty string

mobile/src/main/java/ch/epfl/sdp/mobile/ui/speech_recognition/SpeechFilterRules.kt
```diff
@@ -0,0 +1,55 @@
...
+  /**
+   * Filter rule for Rook
+   * @param speech list of speech' words
+   * @return true if the rule applies to the speech, false otherwise
+   */
+  private fun filterForRook(speech: List<String>): Boolean {
+    return speech.any { sp -> sp.endsOrStartsWithAny(String::endsWith, "ouk", "uk", "uch", "och") }
+  }
+
+  /**
```
KurohanaJuri, line 45: Miss the description

KurohanaJuri (commented):
This is a good idea to have interface for the dictionary. That allow use to extend our recognizer with different language

mobile/src/main/java/ch/epfl/sdp/mobile/ui/speech_recognition/ChessSpeechRecognitionFilter.kt
BadrTad: Good to know !

mobile/src/main/java/ch/epfl/sdp/mobile/ui/speech_recognition/ChessSpeechRecognitionFilter.kt
BadrTad: Yes !

mobile/src/main/java/ch/epfl/sdp/mobile/ui/speech_recognition/ChessSpeechRecognitionFilter.kt
BadrTad: My idea is to at least get some info from the first user unput and then retry. For example if the speech filter recognizer the chess piece and number but not the letter, we retry. 
If the second time, however, we could only recognize the letter but neither the chess piece nor the letter. We could then use the prior info.

mobile/src/main/java/ch/epfl/sdp/mobile/ui/speech_recognition/SpeechFilterRules.kt
BadrTad: Added in `647ee`

mobile/src/main/java/ch/epfl/sdp/mobile/ui/speech_recognition/ChessSpeechRecognitionFilter.kt
KurohanaJuri: This approach didn't seem natural for me. As a user, it's more natural for me to say the whole input again.

I propose to let it as is for the moment and discuss this in the next meeting. 

We can change it according what we decide during the meeting in another PR (It's a small change)

KurohanaJuri (approved):

KurohanaJuri (approved):

mobile/src/main/java/ch/epfl/sdp/mobile/ui/speech_recognition/ChessSpeechRecognitionFilter.kt
BadrTad: Seems good ! I also meant that the user will say the whole input again and hopefully the second time we'll be able to detect the missed word-s and reconstruct the whole move (assuming the player didn't change the move)

mobile/src/main/java/ch/epfl/sdp/mobile/ui/speech_recognition/ChessDictionary.kt
```diff
@@ -0,0 +1,15 @@
+package ch.epfl.sdp.mobile.ui.speech_recognition
+
+/**
+ * Interface representing the vocabulary of chess move commands
+ * @property chessPieces mapping of chess pieces keywords to their homophones
+ * @property numbers mapping of chess number keywords to their homophones
+ * @property letters mapping of chess letter keywords to their homophones
+ * @property placements all possible combinations of chess board placement
```
alexandrepiveteau, line 8: The `@property` KDoc annotation is used for properties declared in the primary constructor. Please document the properties individually instead.

https://kotlinlang.org/docs/kotlin-doc.html

mobile/src/main/java/ch/epfl/sdp/mobile/ui/speech_recognition/ChessDictionary.kt
```diff
@@ -0,0 +1,15 @@
+package ch.epfl.sdp.mobile.ui.speech_recognition
```
alexandrepiveteau, line 1: I think this should go in the application package.

mobile/src/main/java/ch/epfl/sdp/mobile/ui/speech_recognition/ChessDictionary.kt
```diff
@@ -0,0 +1,15 @@
...
+ * @property chessPieces mapping of chess pieces keywords to their homophones
+ * @property numbers mapping of chess number keywords to their homophones
+ * @property letters mapping of chess letter keywords to their homophones
+ * @property placements all possible combinations of chess board placement
+ */
+interface ChessDictionary {
+  val chessPieces: Map<String, List<String>>
+  val numbers: Map<String, List<String>>
+  val letters: Map<String, List<String>>
+  val placements: List<String>
```
alexandrepiveteau, line 14: Question : is the ordering important ? Or should we use some `Set` instead ?

mobile/src/main/java/ch/epfl/sdp/mobile/ui/speech_recognition/ChessSpeechDictionary.kt
```diff
@@ -0,0 +1,73 @@
+package ch.epfl.sdp.mobile.ui.speech_recognition
```
alexandrepiveteau, line 1: Same as above.

mobile/src/main/java/ch/epfl/sdp/mobile/ui/speech_recognition/ChessSpeechRecognitionFilter.kt
```diff
@@ -0,0 +1,210 @@
+package ch.epfl.sdp.mobile.ui.speech_recognition
```
alexandrepiveteau, line 1: Same as above.

mobile/src/main/java/ch/epfl/sdp/mobile/ui/speech_recognition/SpeechRecognitionFilter.kt
```diff
@@ -0,0 +1,14 @@
...
+/** Filtering interface for speech recognizer results */
+interface SpeechRecognitionFilter {
+  /**
+   * @param speeches list of recognized speeches
+   * @param confidencesScores assigned scores to each recognized speech
+   * @return filtered results
+   */
+  fun filterWords(
+      speeches: List<String>,
+      confidencesScores: List<Double>,
```
alexandrepiveteau, line 12: You may want to use a dedicated data class to have a single list of `Result(val text: String, val score: Double)`.

mobile/src/main/java/ch/epfl/sdp/mobile/ui/speech_recognition/SpeechRecognitionFilter.kt
```diff
@@ -0,0 +1,14 @@
...
+  ): String
```
alexandrepiveteau, line 13: What happens if the input lists are empty ? This should maybe return a `String?`

mobile/src/main/java/ch/epfl/sdp/mobile/ui/speech_recognition/ChessSpeechRecognitionFilter.kt
```diff
@@ -0,0 +1,210 @@
...
+   * @return candidate as a chess keyword
+   */
+  private fun bestCandidate(candidates: Map<Word, Double>): String {
+    if (candidates.isEmpty()) {
+      return ""
+    }
+
+    if (candidates.size == 1) {
+      return candidates.entries.first().key
+    }
```
alexandrepiveteau, line 62: I'm not sure we need to special-case this, since the for loop with a single entry seems to have the same result.

mobile/src/main/java/ch/epfl/sdp/mobile/ui/speech_recognition/SpeechFilterRules.kt
```diff
@@ -0,0 +1,58 @@
+package ch.epfl.sdp.mobile.ui.speech_recognition
```
alexandrepiveteau, line 1: Same as above.

mobile/src/main/java/ch/epfl/sdp/mobile/ui/speech_recognition/SpeechRecognitionFilter.kt
```diff
@@ -0,0 +1,14 @@
+package ch.epfl.sdp.mobile.ui.speech_recognition
```
alexandrepiveteau, line 1: Same as above.

mobile/src/test/java/ch/epfl/sdp/mobile/test/ui/speech_recognition/ChessSpeechRecognitionFilterTest.kt
```diff
@@ -0,0 +1,121 @@
+package ch.epfl.sdp.mobile.test.ui.speech_recognition
```
alexandrepiveteau, line 1: Same as above.

alexandrepiveteau (changes requested):

mobile/src/main/java/ch/epfl/sdp/mobile/ui/speech_recognition/ChessDictionary.kt
BadrTad: 👍

mobile/src/main/java/ch/epfl/sdp/mobile/ui/speech_recognition/ChessSpeechRecognitionFilter.kt
BadrTad: Yes indeed !

mobile/src/main/java/ch/epfl/sdp/mobile/ui/speech_recognition/ChessDictionary.kt
BadrTad: No, it was just for convenience, I changed it to Set. Now, the implementation transforms the computed list to a set after doing the mappings on the keys.

mobile/src/main/java/ch/epfl/sdp/mobile/ui/speech_recognition/SpeechRecognitionFilter.kt
BadrTad: Resolved in DMs. We concluded that it's more convenient and less error prone to use nullable string as a return type instead of assuming empty strings for unsuccessful filtering.

KurohanaJuri (approved):

alexandrepiveteau (approved):

PR #271, opened 2022-04-28
Make `Parser` return a `Sequence` of results instead of a `Set`

alexandrepiveteau:
This PR changes the return type of `Parser` to be a `Sequence` rather than a `Set` of results. This way, parser results can be lazily rather than eagerly computed if multiple choices are available.


matt989253 (approved):
Fine

Fouad-sys (approved):
Looks good.

PR #293, opened 2022-05-03
Extract `MatchChessBoardState` from `MatchGameScreenState`

### LINKED ISSUE #262: Create a specific state for `StatefullArScreen.kt`
As said in #251, we need to create a specific statee for `StatefullArScreen.kt`

Citation from the comment
```
 // FIXME : Not the cleanest code, but the faster way to deal with state with different UI. The AR
  //  Screen didn't need all methods in [SnapshotChessBoardState]. The cleaner way is to have a
  //  simpler snapshot that contains only information that will be needed by the
  //  [ArChessBoardScreen]
```

### LINKED ISSUE #286: Refactor chessboard hierarchy

alexandrepiveteau:
This will fix #262 and will fix #286.


KurohanaJuri (approved):
LGTM, thx for this refactoring

Fouad-sys (approved):
Great, thanks!

PR #313, opened 2022-05-07
Tournament Screen Navigation

Fouad-sys:
This PR simply adds a navigation section to the list of tournaments screen (which is empty for now) and adds the functions' and classes' initial declarations for the view and view-model of this screen.


alexandrepiveteau:
@Fouad-sys can we close this PR ?

PR #315, opened 2022-05-07
Fix ambiguity inside `StatefulHomeTest`

### LINKED ISSUE #314: Fix test inside edit profile name
Inside [StatefulHomeTest](https://github.com/epfl-SDP/android/blob/main/mobile/src/androidTest/java/ch/epfl/sdp/mobile/test/state/StatefulHomeTest.kt) the test `given_userIsLoggedIn_when_editProfileName_then_nameShouldBeUpdated` can have different behaviour depending on the device.

We can observe 2 behaviour
- The new string can be appended to the current one
- The new string replaces the existing one

We should enforce the replacement of the input to remove this ambiguity

KurohanaJuri:
Fix #314


alexandrepiveteau (approved):
Thanks !

matt989253 (approved):

PR #325, opened 2022-05-09
Add `Transaction` support to `FirestoreStore` and `FakeStore`

alexandrepiveteau:
This PR adds full support for transactions to `FirestoreStore`, and minimal support for transactions to `FakeStore` (transactions semantics are not respected, but instead document reads and writes are always performed and succeed in order).

### DIFF
--- a/mobile/src/main/java/ch/epfl/sdp/mobile/infrastructure/persistence/store/Store.kt
+++ b/mobile/src/main/java/ch/epfl/sdp/mobile/infrastructure/persistence/store/Store.kt
@@ -1,5 +1,7 @@
 package ch.epfl.sdp.mobile.infrastructure.persistence.store
 
+import kotlin.jvm.Throws
+
 /**
  * An interface representing a place where collections are stored. This is the top-level of the
  * database hierarchy.
@@ -15,9 +17,11 @@ interface Store {
    *
    * This function may throw an [Exception] if a transaction can't be executed.
    *
+   * @throws Exception if a transaction fails.
    * @param R the type of the return value.
    * @param block the body of the [Transaction] to be executed.
    * @return the return value of the transaction.
    */
+  @Throws(Exception::class)
   suspend fun <R> transaction(block: Transaction<DocumentReference>.() -> R): R
 }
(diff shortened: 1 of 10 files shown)


mobile/src/main/java/ch/epfl/sdp/mobile/infrastructure/persistence/store/Store.kt
```diff
@@ -8,4 +8,16 @@ interface Store {
 
   /** Accesses a collection with the given [String] identifier. */
   fun collection(path: String): CollectionReference
+
+  /**
+   * Runs the give [Transaction] on the [Store], executing all the operations atomically. A
+   * [Transaction] must perform all its reads before performing all of its writes.
+   *
+   * This function may throw an [Exception] if a transaction can't be executed.
```
KurohanaJuri, line 16: Can we add this information with a keyword, something like `@Exception` ? I don't if it's possible

mobile/src/main/java/ch/epfl/sdp/mobile/infrastructure/persistence/store/Transaction.kt
```diff
@@ -0,0 +1,81 @@
+package ch.epfl.sdp.mobile.infrastructure.persistence.store
+
+import kotlin.reflect.KClass
+
+/**
+ * A [Transaction] offers a way for clients to perform multiple [get] of documents, followed by some
+ * [set] and [update]. At most 500 documents may be modified per [Transaction].
```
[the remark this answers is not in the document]
alexandrepiveteau: Yes.

mobile/src/main/java/ch/epfl/sdp/mobile/infrastructure/persistence/store/Store.kt
alexandrepiveteau: Fixed in d1b4b2e0ff48edc3a1438212e25d1472214c6741.

matt989253 (approved):
Looks great to me! As discussed, we'll have to be careful with the `@UnsafeVariance`.

PR #346, opened 2022-05-12
Tournament creation time

Fouad-sys:


mobile/src/main/java/ch/epfl/sdp/mobile/ui/i18n/English.kt
```diff
@@ -184,3 +185,15 @@ object English : LocalizedStrings {
...
+  if (this.toString().contains("d")) {
+    return "${this.toLong(DurationUnit.DAYS)} days"
+  } else if (this.toString().contains("h")) {
+    return "${this.toLong(DurationUnit.HOURS)} hours"
+  } else if (this.toString().contains("m")) {
+    return "${this.toLong(DurationUnit.MINUTES)} minutes"
+  } else {
+    return "${this.toLong(DurationUnit.SECONDS)} seconds"
+  }
+}
```
alexandrepiveteau, line 199: You probably should not use the `toString()` representation of the `Duration` to transform it to some English text.What would you think of inspecting the inner fields directly ?

mobile/src/main/java/ch/epfl/sdp/mobile/ui/i18n/English.kt
```diff
@@ -184,3 +185,15 @@ object English : LocalizedStrings {
   override val tournamentsCreateElimDemomN = { denominator: Int -> "1 / $denominator" }
   override val tournamentsCreateElimDepthFinal = "Final"
 }
+
+fun Duration.toDurationString(): String {
```
alexandrepiveteau, line 189: Missing some documentation, and this must be private. Also, I'd rename it to `toEnglishString`, to avoid potential name clashes when new languages are added.

mobile/src/androidTest/java/ch/epfl/sdp/mobile/test/state/StatefulContestScreenTest.kt
```diff
@@ -88,4 +88,111 @@ class StatefulContestScreenTest {
...
+              TournamentDocument(
+                  "tid1", "2", "Tournament 1", playerIds = listOf("1"), creationTime = startTime))
+        }
+      }
+
+      val (_, _, strings) = rule.setContentWithTestEnvironment(store = store) { StatefulHome(user) }
+
+      rule.onNodeWithText(strings.sectionSocial).performClick()
+      val duration = 1.minutes
+      Thread.sleep(duration.toLong(DurationUnit.MILLISECONDS))
```
alexandrepiveteau, line 110: You should never call `Thread.sleep` in a test.

mobile/src/androidTest/java/ch/epfl/sdp/mobile/test/state/StatefulContestScreenTest.kt
```diff
@@ -88,4 +88,111 @@ class StatefulContestScreenTest {
...
+      rule.onNodeWithText(strings.tournamentsCreateQualifierSize0).performClick()
+      rule.onNodeWithText(strings.tournamentsCreateElimDemomN(2)).performClick()
+
+      rule.onNodeWithText(strings.tournamentsCreateActionCreate).performClick()
+      rule.onNodeWithContentDescription(strings.tournamentDetailsBackContentDescription)
+          .performClick()
+      rule.onNodeWithText(strings.sectionSocial).performClick()
+
+      val duration = 1.minutes
+      Thread.sleep(duration.toLong(DurationUnit.MILLISECONDS))
```
alexandrepiveteau, line 137: Same as above.

mobile/src/main/java/ch/epfl/sdp/mobile/application/tournaments/TournamentFacade.kt
```diff
@@ -63,6 +63,7 @@ class TournamentFacade(private val auth: Auth, private val store: Store) {
       user: AuthenticatedUser,
       name: String,
       maxPlayers: Int,
+      creationTime: Long,
```
alexandrepiveteau, line 66: It's probably better to "hide" the creationTime from the caller and let the `TournamentFacade` populate it, since creating a tournament should not be done in the future or in the past.

mobile/src/main/java/ch/epfl/sdp/mobile/application/StoreDocuments.kt
```diff
@@ -114,6 +115,7 @@ data class TournamentDocument(
     val adminId: String? = null,
     val name: String? = null,
     val maxPlayers: Int? = null,
+    val creationTime: Long? = null,
```
alexandrepiveteau, line 118: Maybe worth specifying the unit there. What would you think of naming this `creationTimeEpochMillis` for instance ?

mobile/src/main/java/ch/epfl/sdp/mobile/state/StatefulContestsScreen.kt
```diff
@@ -34,7 +40,27 @@ data class TournamentAdapter(val tournament: Tournament, val currentUser: Authen
...
+
+  if (duration >= 1.days) {
+    return duration.inWholeDays.days
+  } else if (duration >= 1.hours) {
+    return duration.inWholeHours.hours
+  } else if (duration >= 1.minutes) {
+    return duration.inWholeMinutes.minutes
+  }
+
+  return duration.inWholeSeconds.seconds
```
alexandrepiveteau, line 63: Question : what would you think of letting the facade provide the duration since the tournament was open ? These computations seem like something that's domain specific.

alexandrepiveteau (changes requested):

mobile/src/main/java/ch/epfl/sdp/mobile/ui/i18n/English.kt
Fouad-sys: Hey! Yes that was my initial approach but I couldn't access the inner fields actually. Are you aware of a way to do so?

mobile/src/main/java/ch/epfl/sdp/mobile/state/StatefulContestsScreen.kt
Fouad-sys: This was an approach I had actually, but it would require adding the facade as an additional argument to the TournamentAdapter correct?

alexandrepiveteau: Not really; you could provide the duration in the tournament objects returned by the facade.

mobile/src/main/java/ch/epfl/sdp/mobile/application/tournaments/Tournament.kt
```diff
@@ -80,7 +86,8 @@ interface Tournament {
  * @receiver the [TournamentDocument] that we're transforming.
  * @param user the [AuthenticatedUser] that we see this [Tournament] as.
  * @param store the [Store] used to perform changes.
+ * @param time the [Time] used to calculate the duration of creation of the tournament.
  * @return the [Tournament] instance.
  */
-fun TournamentDocument.toTournament(user: AuthenticatedUser, store: Store): Tournament =
-    StoreDocumentTournament(this, user, store)
+fun TournamentDocument.toTournament(user: AuthenticatedUser, store: Store, time: Time = SystemTime()): Tournament =
```
alexandrepiveteau, line 92:
```suggestion
fun TournamentDocument.toTournament(user: AuthenticatedUser, store: Store, time: Time): Tournament =
```

mobile/src/main/java/ch/epfl/sdp/mobile/ui/i18n/English.kt
```diff
@@ -191,3 +192,16 @@ object English : LocalizedStrings {
...
+  if (this.toString().contains("d")) {
+    return "${this.toLong(DurationUnit.DAYS)} days"
+  } else if (this.toString().contains("h")) {
+    return "${this.toLong(DurationUnit.HOURS)} hours"
+  } else if (this.toString().contains("m")) {
+    return "${this.toLong(DurationUnit.MINUTES)} minutes"
+  } else {
+    return "${this.toLong(DurationUnit.SECONDS)} seconds"
+  }
+}
```
alexandrepiveteau, line 207: You may not want to match on the string representation of the `Duration`. Shouldn't you see if it's above a certain threshold, and then use the `inWholeDays` and co. methods to retrieve the duration ?

Also, what would you say of performing the rounding of durations here directly ?

alexandrepiveteau (commented):
Looking very nice ! There are a few nitpicks to handle, but the overall structure LGTM.

Just a question - what would you think of defaulting to `FakeTimeProvider` for all the tests, rather than `SystemTimeProvider` ? I feel like a deterministic provider is a better default.

mobile/src/main/java/ch/epfl/sdp/mobile/ui/i18n/English.kt
```diff
@@ -191,3 +194,19 @@ object English : LocalizedStrings {
...
+private fun Duration.toEnglishString(): String {
+  if (this >= 1.days) {
+    return "${this.inWholeDays} days"
+  } else if (this >= 1.hours) {
+    return "${this.inWholeHours} hours"
+  } else if (this >= 1.minutes) {
+    return "${this.inWholeMinutes} minutes"
+  }
+
+  return "${this.inWholeSeconds} seconds"
```
KurohanaJuri, line 211: Why the return "${this.inWholeSeconds} seconds" isn't in an else branch ? Do you have a particular reason to do that ?

KurohanaJuri, line 211: Why the return `"${this.inWholeSeconds} seconds"` isn't in an else branch ? Do you have a particular reason to do that ?

mobile/src/main/java/ch/epfl/sdp/mobile/application/tournaments/StoreDocumentTournament.kt
```diff
@@ -177,4 +188,28 @@ class StoreDocumentTournament(
       }
     }
   }
+
+  /**
+   * Obtains the elapsed duration from the startTime and rounds it to the closest unit of time
```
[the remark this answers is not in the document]
Fouad-sys: Fixed in cfeb7b2726ea8c78c7fcc5a4ba2e5be8ac91d7db

mobile/src/main/java/ch/epfl/sdp/mobile/ui/i18n/English.kt
Fouad-sys: No actually, just for having a default.

mobile/src/androidTest/java/ch/epfl/sdp/mobile/test/state/Dsl.kt
```diff
@@ -103,7 +104,7 @@ suspend fun ComposeContentTestRule.setContentWithTestEnvironment(
   val socialFacade = SocialFacade(auth, store)
   val chessFacade = ChessFacade(auth, store, assets)
   val speechFacade = SpeechFacade(recognizer)
-  val tournamentFacade = TournamentFacade(auth, dataStoreFactory, store)
+  val tournamentFacade = TournamentFacade(auth, dataStoreFactory, store, SystemTimeProvider)
```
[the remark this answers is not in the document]
Fouad-sys: Fixed in 16a0ff3b778223d403a8e858260383879e31ddba

mobile/src/androidTest/java/ch/epfl/sdp/mobile/test/infrastructure/persistence/store/fake/FakeTimeProvider.kt
```diff
@@ -0,0 +1,22 @@
+package ch.epfl.sdp.mobile.test.infrastructure.persistence.store.fake
```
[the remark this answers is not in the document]
Fouad-sys: Fixed in https://github.com/epfl-SDP/android/commit/16a0ff3b778223d403a8e858260383879e31ddba

BadrTad (approved):

KurohanaJuri (approved):

mobile/src/androidTest/java/ch/epfl/sdp/mobile/test/state/StatefulContestScreenTest.kt
```diff
@@ -91,4 +96,149 @@ class StatefulContestScreenTest {
...
+
+      val time = FakeTimeProvider
+      val assets = emptyAssets()
+      val authFacade = AuthenticationFacade(auth, store)
+      val socialFacade = SocialFacade(auth, store)
+      val chessFacade = ChessFacade(auth, store, assets)
+      val speechFacade = SpeechFacade(FailingSpeechRecognizerFactory)
+      val tournamentFacade = TournamentFacade(auth, dataStoreFactory, store, time)
+
+      val user = authFacade.currentUser.filterIsInstance<AuthenticatedUser>().first()
```
alexandrepiveteau, line 127: You never logged in, but actually can still fetch the `currentUser`. This looks like a bug in `FakeStore` which we can fix later.

alexandrepiveteau (approved):

PR #358, opened 2022-05-14
Add modifier on the last move

### LINKED ISSUE #357: Display the last played move

KurohanaJuri:
The player can see the last action that was performed, the cell have another colour with animated dashed lane.

Closes #357 

image

### DIFF
--- a/mobile/src/androidTest/java/ch/epfl/sdp/mobile/test/ui/game/ChessBoardTest.kt
+++ b/mobile/src/androidTest/java/ch/epfl/sdp/mobile/test/ui/game/ChessBoardTest.kt
@@ -53,6 +53,7 @@ class ChessBoardTest {
     }
 
     override fun onPositionClick(position: Position) = Unit
+    override val lastMove: Position? = null
   }
 
   @Test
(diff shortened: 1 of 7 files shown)


matt989253 (commented):
Great work, it looks nice!

However, what would you think about removing the dashed line and also marking the "from" position, like this? This is usually how it is done on other chess apps, and I think it gives the user more useful information! (The dashed line is more subjective, but I think it looks a litttle bit "heavy"

image

KurohanaJuri:
> However, what would you think about removing the dashed line and also marking the "from" position, like this? This is usually how it is done on other chess apps, and I think it gives the user more useful information! (The dashed line is more subjective, but I think it looks a litttle bit "heavy", especially when the user has another selected square)

I agree that the dashed line is a bit heavy, but I put this to distinguish the last move and the checkmate. I make your suggestion, do you think that looks good ? The colours are too similar to me, I don't know what colour to put. I don't think that green is a good idea. 

image

matt989253:
Although they are indeed quite similar, I don't think this is a problem. First, they will always be two squares lit up from the previous move (except for the first move), so in the event of a check position appearing too there should not be any confusion as to which is which.

Also, it is impossible for a king to both be in check and have played the last move (rules of chess prevent this), so there also cannot be any confusion because of that. 

A way many chess apps show a king in check is by using a diffuse/blurry red circle behind the king instead of coloring the whole square. This could be an option, but I don't think it's worth the effort.

All in all, I think the way it looks right now is the best solution.

matt989253 (approved):

alexandrepiveteau (commented):
Looks great ! I think the doc might reflect the previous behavior with dashed borders, the rest looks good.

alexandrepiveteau (approved):

matt989253 (approved):

PR #345, opened 2022-05-12
Implement pool and direct elimination states

alexandrepiveteau:
This PR adds Statefulness to Tournaments.
It adds:
- Starting the pool phase as admin
- Advancing a pool's best of round as admin
- Starting direct elimination rounds as admin
- advancing elimination rounds as admin
- spectating a direct elimination round

PR #378, opened 2022-05-14
Optimize `Query.get()` to avoid `asFlow()` usage

alexandrepiveteau:
This fixes (an ugly) shortcut that was taken during #345.


matt989253 (approved):

BadrTad (approved):
Looks good to me !

PR #385, opened 2022-05-16
Use `setContentWithTestEnvironment` in existing tests

alexandrepiveteau:
In order to simplify the addition of the future `DataStore<Preferences>` in the `TournamentFacade`, many tests are migrated to use `setContentWithTestEnvironment`, which will make adding more dependencies to the facades considerably easier.


BadrTad (approved):
Thanks ! Tests looks much better. Also I see that this was done to provide empty facades only could it be extended to support custom facades  like non-empty stores, auths...  ?

alexandrepiveteau:
> Thanks ! Tests looks much better. Also I see that this was done to provide empty facades only could it be extended to support custom facades like non-empty stores, auths... ?

Actual, the signature of `setContentWithTestEnvironment` is as follows :

```kotlin
suspend fun ComposeContentTestRule.setContentWithTestEnvironment(
    store: Store = emptyStore(),
    auth: Auth = emptyAuth(),
    assets: AssetManager = emptyAssets(),
    recognizer: SpeechRecognizerFactory = FailingSpeechRecognizerFactory,
    strings: LocalizedStrings = English,
    content: @Composable TestEnvironment.() -> Unit,
): TestEnvironment {
```

By default, empty stores, auth and assets are provided, but a client could also call the method with a populated `Store`, `Auth` or whatever which will be re-used across the facades.

Fouad-sys (approved):
Thank you for this heavy refactoring Alexandre. Really simplifies future tests. Greatly done!

PR #410, opened 2022-05-24
Remove the visible divider and improve indicator behaviour on scroll

alexandrepiveteau:
This PR slightly tweaks the visuals of the tab bar of the contests screen for the video. It also adds some simple animations when contests are re-ordered.

### DIFF
--- a/mobile/src/main/java/ch/epfl/sdp/mobile/ui/tournaments/TournamentDetails.kt
+++ b/mobile/src/main/java/ch/epfl/sdp/mobile/ui/tournaments/TournamentDetails.kt
@@ -294,6 +294,7 @@ private fun DetailsTopBar(
             selectedTabIndex = selected,
             backgroundColor = MaterialTheme.colors.background,
             indicator = {}, // Hide the default indicator.
+            divider = {}, // Hide the default divider.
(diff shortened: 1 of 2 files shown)


KurohanaJuri (approved):

mobile/src/main/java/ch/epfl/sdp/mobile/ui/tournaments/TournamentDetails.kt
```diff
@@ -507,3 +510,14 @@ private fun DetailsMatch(
     }
   }
 }
+
+/**
+ * A variation of [PagerState.currentPage] which displays the page which is the "most visible" on
```
Fouad-sys, line 515: [Question] By "most visible" do you mean the one in the frontmost view?

Fouad-sys (approved):
LGTM

mobile/src/main/java/ch/epfl/sdp/mobile/ui/tournaments/TournamentDetails.kt
alexandrepiveteau: The one whose visible-width to total-width ratio is the highest.

PR #415, opened 2022-05-25
Add a smooth displacement on pieces

### LINKED ISSUE #369: Animate the chess piece moves in AR

KurohanaJuri:
Implement a smooth displacement when the piece changes position. The displacement move first up, to the target position, and then down. 

We have a small bug, but we have a small workaround (See below to have more information). 

Close #369

----
## Workaround

For some reason, if we add the board's node to the scene via a function, the animation doesn't show up. 
The workaround is to add the board's "by default" when the scene is created and set the visibility to false. When the user click on the screen, we set the visibility to true.

### DIFF
--- a/mobile/src/main/java/ch/epfl/sdp/mobile/ui/game/ar/ArChessBoardScreen.kt
+++ b/mobile/src/main/java/ch/epfl/sdp/mobile/ui/game/ar/ArChessBoardScreen.kt
@@ -53,8 +53,16 @@ fun <Piece : ChessBoardState.Piece> ArChessBoardScreen(
               // Place the chess board on the tapped position.
               arSceneView.onTouchAr =
                   { hitResult, _ ->
-                    anchorOrMoveBoard(arSceneView, chessScene, hitResult.createAnchor())
+                    anchorOrMoveBoard(/*arSceneView,*/ chessScene, hitResult.createAnchor())
                   }
+
+              /**
+               * FIXME : Workaround : A strange bug make the animation fail when we add the child
+               * via a function. To solve this quickly, we add the board when is loaded and set it
+               * to invisible. We reset the visibility when the user tap on the screen
+               */
+              arSceneView.addChild(this.boardNode)
+              this.boardNode.isVisible = false
             }
 
         arSceneView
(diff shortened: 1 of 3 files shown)


mobile/src/main/java/ch/epfl/sdp/mobile/ui/game/ar/ArChessBoardScreen.kt
```diff
@@ -68,22 +65,60 @@ fun <Piece : ChessBoardState.Piece> ArChessBoardScreen(
...
-    arSceneView: ArSceneView,
-    chessScene: ChessScene<Piece>?,
+private fun anchorOrMoveBoard(
+    // arSceneView: ArSceneView,
+    boardNode: ArModelNode,
     anchor: Anchor
 ) {
 
-  val currentChessScene = chessScene ?: return
+  // FIXME : Workaround see line 55
```
alexandrepiveteau, line 77: I don't think this is the right line number.

```suggestion
  // FIXME : Workaround see line 117
```

mobile/src/main/java/ch/epfl/sdp/mobile/ui/game/ar/ArChessBoardScreen.kt
```diff
@@ -68,22 +65,60 @@ fun <Piece : ChessBoardState.Piece> ArChessBoardScreen(
...
+  }*/
+
+  if (!boardNode.isVisible) boardNode.isVisible = true
+
+  boardNode.anchor = anchor
+}
+
+/**
+ * Create an instance of [ChessScene] and setup the onTouch callback
+ *
```
alexandrepiveteau, line 91: Missing documentation for `Piece` type parameter.

mobile/src/main/java/ch/epfl/sdp/mobile/ui/game/ar/ArChessBoardScreen.kt
KurohanaJuri: Yeah, I forget to change it after my refactoring ^^"

barmettlerl (approved):
LGTM

mobile/src/main/java/ch/epfl/sdp/mobile/ui/game/ar/ArChessBoardScreen.kt
KurohanaJuri: Fix in https://github.com/epfl-SDP/android/pull/415/commits/3c79295548979f5ffd972831dafda98d86b50618

matt989253 (approved):
Indeed, it's very good looking!

PR #418, opened 2022-05-26
Fix `FakeDataStoreFactory`

alexandrepiveteau:
Due to the factory being an object, a single instance was reused across all the tests. This leads to cross-test interactions, which may make some valid tests fail.


matt989253 (approved):
Thanks for the bugfix!

barmettlerl (approved):

PR #427, opened 2022-05-27
Trim white spaces in AuthScreen

BadrTad:
Trim white spaces for email and name text fields.

### DIFF
--- a/mobile/src/main/java/ch/epfl/sdp/mobile/ui/authentication/AuthenticationScreen.kt
+++ b/mobile/src/main/java/ch/epfl/sdp/mobile/ui/authentication/AuthenticationScreen.kt
@@ -60,7 +60,7 @@ fun AuthenticationScreen(
       Column {
         TextField(
             value = state.email,
-            onValueChange = { state.email = it },
+            onValueChange = { state.email = it.trim() },
             label = { Text(strings.authenticateEmailHint) },
             modifier = Modifier.fillMaxWidth(),
             keyboardOptions = KeyboardOptions(keyboardType = KeyboardType.Email),
@@ -72,7 +72,7 @@ fun AuthenticationScreen(
         ) {
           TextField(
               value = state.name,
-              onValueChange = { state.name = it },
+              onValueChange = { state.name = it.trim() },
               label = { Text(strings.authenticateNameHint) },
               modifier = Modifier.fillMaxWidth().padding(top = 16.dp),
           )


mobile/src/main/java/ch/epfl/sdp/mobile/ui/authentication/AuthenticationScreen.kt
```diff
@@ -72,7 +72,7 @@ fun AuthenticationScreen(
         ) {
           TextField(
               value = state.name,
-              onValueChange = { state.name = it },
+              onValueChange = { state.name = it.trim() },
```
KurohanaJuri, line 75: With this, we cannot have a username with a space inside (like "Chau Ying")

BadrTad: Not really, trim only removes leading and trailing white spaces, it's actually equivalent to `name.trimStart().trimEnd()`. So `"Chau Ying".trim()` will result in `"Chau Ying"` as expected and no words will be omitted.

matt989253: But if i'm not mistaked, since it is `onValueChange`, you will never be able to write the ` space` character, except if you write the name without spaces and then click where you wanted to insert it initially.

alexandrepiveteau: @matt989253 and @KurohanaJuri are right - applying the trim in `onValueChange` restricts what users can type; trimming the names should be applied only when submitting the values and thus creating the account, and could (for instance) be performed in the facade directly.

BadrTad: I would prefer to trim the inputs when making the calls to the concerned facade methods in the stateful authentication screen.

alexandrepiveteau:
Question : is there any reason for which the trimming happens in the stateful composable, and not in the facade instead ? I feel like implementing the former would make the trimming more reliable.

Additionally, making it part of the facade would help write 1-2 tests which validate the trimming behaviour.

alexandrepiveteau (commented):
Question : is there any reason for which the trimming happens in the stateful composable, and not in the facade instead ? I feel like implementing the former would make the trimming more reliable.

Additionally, making it part of the facade would help write 1-2 tests which validate the trimming behaviour.

matt989253 (dismissed):

alexandrepiveteau (dismissed):

mobile/src/main/java/ch/epfl/sdp/mobile/application/authentication/AuthenticationFacade.kt
```diff
@@ -57,7 +57,7 @@ class AuthenticationFacade(private val auth: Auth, private val store: Store) {
   suspend fun signInWithEmail(
       email: String,
       password: String,
-  ): AuthenticationResult = authenticate { auth.signInWithEmail(email, password) }
+  ): AuthenticationResult = authenticate { auth.signInWithEmail(email.trim(), password.trim()) }
```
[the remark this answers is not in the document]
BadrTad: Ohh my bad silly mistake. I did my best to keep in mind that passwords should not be trimmed but tiredness got me last night.

alexandrepiveteau (approved):

matt989253 (approved):

PR #441, opened 2022-05-31
Work around `HomeActivity` failing test

alexandrepiveteau:
This PR works around the bug in `Mockk` issue which makes static function calls unusable after they've been mocked once.

Additionally, it fixes some unmock calls missing for some of the existing tests.


matt989253 (approved):
LGTM

PR #444, opened 2022-05-31
Introduce Detekt to perform some static analysis of the code and comments

alexandrepiveteau:
This PR introduces Detekt, a static analysis tool for Kotlin, and configures it to enforce that all the public properties, classes and functions are documented in the codebase. This is a good way to enforce that the submission will be fully documented, and that we won't have forgotten a single class / function / property.

Additionally, it sets up a GitHub Actions script to make sure these rules are enforced for all contributions.

### DIFF
--- a/.github/workflows/detekt.yml
+++ b/.github/workflows/detekt.yml
@@ -1,0 +1,13 @@
+name: Run detekt
+
+on: [ push, pull_request ]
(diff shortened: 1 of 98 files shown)


BadrTad (approved):
Great addition 👍

PR #456, opened 2022-06-06
Strengthen en-passant rules constaints for puzzles

alexandrepiveteau:
This PR fixes a problem I encountered while playing some of the provided puzzles.

In some rare cases, the en-passant behaviour could be proposed to pawns in puzzles when they were not located on the right row. This caused a problem for some specific puzzles if there was a piece located right behind the pawn which could be taken en-passant, since multiple `Move` would then be available for the pawn (both the en-passant, or the take of the piece behind). As a reminder, in case of ambiguous moves which are not promotions, the viewmodel simply drops the user action since it assumes there was an error in the engine.

This patch strengthens the en-passant conditions to make sure the ambiguous cases are not possible anymore, by enforcing that en-passant happens on a specific row, and that the target position of the "taking" pawn is free.


BadrTad:
LGTM Thanks !

PR #457, opened 2022-06-06
Improve scrolling performance in `UserScreen`

alexandrepiveteau:
This PR improves scrolling performance in the list of matches and puzzles for profiles. This is done in two ways :

- the `LazyColumn` state is no longer read in the composition but rather as a `derivedStateOf`, reducing and narrowing the scope of recompositions to animate the elevation of the toolbars;
- the matches and puzzles are now keyed, which may help with item re-using.

### DIFF
--- a/mobile/src/main/java/ch/epfl/sdp/mobile/ui/profile/ProfileScreen.kt
+++ b/mobile/src/main/java/ch/epfl/sdp/mobile/ui/profile/ProfileScreen.kt
@@ -27,12 +27,16 @@ import ch.epfl.sdp.mobile.ui.social.ChessMatch
  * @param state state of the ProfileScreen.
  * @param modifier the [Modifier] for this composable.
  * @param contentPadding the [PaddingValues] to apply to this screen.
+ * @param matchKey the key for each match item.
+ * @param puzzleKey the key for each puzzle item.
  */
 @Composable
 fun <C : ChessMatch, P : PuzzleInfo> ProfileScreen(
     state: VisitedProfileScreenState<C, P>,
     modifier: Modifier = Modifier,
     contentPadding: PaddingValues = PaddingValues(),
+    matchKey: ((C) -> Any)? = null,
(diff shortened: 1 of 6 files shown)


BadrTad (approved):

BadrTad:
LGTM Thanks !

mobile/src/main/java/ch/epfl/sdp/mobile/ui/profile/ProfileScreen.kt
```diff
@@ -26,25 +27,30 @@ import ch.epfl.sdp.mobile.ui.social.ChessMatch
...
     contentPadding: PaddingValues = PaddingValues(),
+    matchKey: ((C) -> Any)? = null,
+    puzzleKey: ((P) -> Any)? = null,
 ) {
   val lazyColumnState = rememberLazyListState()
   val tabBarState = rememberProfileTabBarState(state.pastGamesCount, state.solvedPuzzlesCount)
-  val targetElevation = if (lazyColumnState.firstVisibleItemIndex >= 1) 4.dp else 0.dp
-  val elevation by animateDpAsState(targetElevation)
+  val targetElevation by remember {
+    derivedStateOf { if (lazyColumnState.firstVisibleItemIndex >= 1) 4.dp else 0.dp }
```
barmettlerl, line 44: Why having this if statement here -> this doesn't lead to an ugly animation jump ?

barmettlerl (approved):

mobile/src/main/java/ch/epfl/sdp/mobile/ui/profile/ProfileScreen.kt
alexandrepiveteau: The `animateDpAsState` call was moved inside the toolbar, so `targetElevation` changes are properly animated.

However, because the `lazyColumnState.firstVisibleItemIndex` value is read within a remembered `derivedStateOf`, we don't recompose each time the value changes and the scrolling is much smoother.

